PR #2, opened 2018-03-09
All base objects are created - foundation is ready.

indradeep:


blurr/core/base.py
```diff
@@ -0,0 +1,283 @@
...
+                                        'Schema definition must specify a list of {}.'.format(
+                                            self._nested_item_attribute))
+
+        # Ensure that the nested schema attribute contains a list of one or more items
+        if len(spec[self._nested_item_attribute]) == 0:
+            self.raise_validation_error(spec, self._nested_item_attribute,
+                                        'Schema definition must have at least one item under {}.'.format(
+                                            self._nested_item_attribute))
+
+    def load(self, spec: Dict[str, Any]) -> None:
```
vigbk: whats the reason the make load and validate as public if they are always called internally?

blurr/core/transformer.py
```diff
@@ -0,0 +1,48 @@
...
+    All Transformer Schema inherit from this base.  Adds support for handling
+    the required attributes of a schema.
+    """
+
+    ATTRIBUTE_VERSION = 'Version'
+    ATTRIBUTE_DESCRIPTION = 'Description'
+    ATTRIBUTE_DATA_GROUPS = 'DataGroups'
+
+    def __init__(self, spec: Dict[str, Any]) -> None:
+        super().__init__(spec, self.ATTRIBUTE_DATA_GROUPS)
```
vigbk: Does this mean you can have only one collection in a given schema. For ex how will it handle storage and datagroups ?

mserranom:
I'd include 2 changes:

1. Add the circleci build descriptor, [check the descriptor in master](https://github.com/productml/blurr/blob/master/.circleci/config.yml), so we have a successful build

2. Move `tests` from the parent folder inside `blurr`, and restructure files accordingly:

From:
```
.
├── README.md
├── Pipfile
├── Pipfile.lock
├── blurr
│   ├── core
│   └── runner
└── tests
    ├── core
    └── runner

```

To:
```
.
├── README.md
├── Pipfile
├── Pipfile.lock
├── blurr
│   ├── blurr
│   │    ├── core
│   │    └── runner
│   ├── tests
│   │    ├── core
│   │    └── runner
│   ├── Pipfile
│   └── Pipfile.lock
└── tests
```

The reason is that we'll need a place to store acceptance tests, and those have different `Pipfile` than the root project. Inside `blurr` we'll have only unit tests, which are an inherent part of the module.

An alternative would be to create a separate folder for the acceptance tests:
```
.
├── README.md
├── Pipfile
├── Pipfile.lock
├── blurr
│   ├── core
│   └── runner
├── acceptance-test
│   ├── Pipfile
│   └── Pipfile.lock
└── tests
    ├── core
    └── runner

```

But this results in a very unpredictable layout,  with anidated `Pipfile` which is a bad practice.

blurr/core/transformer.py
indradeep: There is one primary tree - which is the hierarchy of the rollup generated - with the identity as the root.  DataGroup is part of this hierarchy.

Storage is an external entity, used for execution, and is part of neither the schema object, nor the transformation storage. They will be handled separately.

blurr/core/base.py
indradeep: These are protected methods.  I'm still not sure whether to keep them public or not... perhaps private is a better option.
I'll make the changes in the upcoming PRs.

indradeep:
I and Prashast are both on the side of the `acceptance-test` folder.  Let us resolve the PR and create a branch from `core` to add the unit testing and CI stuff.

PR #16, opened 2018-03-14
added build script and circleci file

mserranom:
* Currently tests are broken!!
* mypy will fail either, it's commented out

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -18,7 +18,7 @@ pipenv run yapf -i -r . --style='{allow_split_before_dict_value : false}'
 
 echo "running tests..."
 export PYTHONPATH=`pwd`:$PYTHONPATH
-pipenv run pytest tests/*/*
+pipenv run pytest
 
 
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -0,0 +1,24 @@
...
+
+echo "formatting and validating code..."
+pipenv run yapf -i -r . --style='{allow_split_before_dict_value : false}'
+# TODO: fix mypy errors
+# pipenv run mypy blurr/*/*.py --ignore-missing-imports --disallow-untyped-defs --disallow-untyped-calls
+
+
+echo "running tests..."
+export PYTHONPATH=`pwd`:$PYTHONPATH
+pipenv run pytest tests/*/*
```
prashastk: you can just do `pipenv run pytest` now with the fix in https://github.com/productml/blurr/pull/17

mserranom: yay!

PR #18, opened 2018-03-14
YAML validation with Yamale lib

mserranom:
This is a proof of concept of validation with Yamale library. It's not intented to merge, but potentially we can migrate all validations following this branch if we fancy.

I think it's a very powerful tool, this approach is quite clean and less verbose, and supports reusing easily custom validations.

Check my inline comments and please provide some feedback

### DIFF
--- a/blurr/core/base.py
+++ b/blurr/core/base.py
@@ -17,19 +17,11 @@ class DTCType(Validator):
     # all valid types would go here, I just made this up
     VALUES = list(SCHEMA_MAP.keys()) + ['ProductML:DTC:Streaming']
 
-    __name__ = "DTC Type"
-
     def _is_valid(self, value):
         return value in self.VALUES
 
-    def __str__(self):
-        return "DTC Valid Type"
-
-    def __repr__(self):
-        return self.__str__()
-
     def get_name(self):
-        return self.__str__()
+        return "DTC Valid Type"
 
 
 class BaseSchema(ABC):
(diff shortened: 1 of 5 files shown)


tests/runner/local_runner_test.py
```diff
@@ -2,7 +2,8 @@
 
 
 def test_local_runner():
-    local_runner = LocalRunner(['tests/runner/data/raw.json'], '', 'tests/runner/data/sample.yml')
```
mserranom: this is just a yapf formatting change

blurr/core/base.py
```diff
@@ -17,12 +35,24 @@ class BaseSchema(ABC):
     ATTRIBUTE_TYPE = 'Type'
     ATTRIBUTE_WHEN = 'When'
 
+    SCHEMA = '''
+        Name: str(min=1)
+        Type: dtc_type()
```
indradeep: Type t different levels have different meanings.  The typeloader that loads the type into the system will validate if the type name is valid.  

I'm not sure that we want to define the DTC type in the base.

blurr/core/base.py
```diff
@@ -1,9 +1,27 @@
...
-from blurr.core.loader import TypeLoader
+from blurr.core.loader import TypeLoader, SCHEMA_MAP
+
+
+# This would be in a more central place along with other validation code
+class DTCType(Validator):
+    TAG = 'dtc_type'
+
+    # all valid types would go here, I just made this up
+    VALUES = list(SCHEMA_MAP.keys()) + ['ProductML:DTC:Streaming']
```
indradeep: This layer is not aware of DTC / Streaming.

mserranom:
closing, I'll raise a separate PR with the production-quality code

PR #25, opened 2018-03-15
Yaml validation with yamale lib

mserranom:
This PR is quite simple, but it has too many lines of code. 

Follow this steps and the links to review:

1. I created 2 schemas for [window](https://github.com/productml/blurr/pull/25/files#diff-d6887fb5edf872d26fbdcb075977522e) and [streaming](https://github.com/productml/blurr/pull/25/files#diff-2c67ec53b7caa90910ec3fa170d970c8) DTCs, they define the syntax of the DTCs and are the **single source of truth**
2. The function to trigger a validation, and the validators for custom data types are in [schema_validator.py](https://github.com/productml/blurr/pull/25/files#diff-28edbf32ab28bc0b6f449228c5187fe2)
3. I added many test examples, each DTC in a single file, but all the test cases are together in [schema_validator_test.py](https://github.com/productml/blurr/pull/25/files#diff-3513c8e46a31d25cb77a469df78c2232)
4. I updated [local_runner.py](https://github.com/productml/blurr/pull/25/files#diff-c901f142e80ee676edc3757ab8f20f19) to perform a validation.
5. [BaseSchema](https://github.com/productml/blurr/pull/25/files#diff-df330e71e0842d8336f4280a7b0d54bc) doesn't need to do certain validations anymore, since they're performed previously with the syntax check

### DIFF
--- a/blurr/core/base.py
+++ b/blurr/core/base.py
@@ -17,19 +17,11 @@ class DTCType(Validator):
     # all valid types would go here, I just made this up
     VALUES = list(SCHEMA_MAP.keys()) + ['ProductML:DTC:Streaming']
 
-    __name__ = "DTC Type"
-
     def _is_valid(self, value):
         return value in self.VALUES
 
-    def __str__(self):
-        return "DTC Valid Type"
-
-    def __repr__(self):
-        return self.__str__()
-
     def get_name(self):
-        return self.__str__()
+        return "DTC Valid Type"
 
 
 class BaseSchema(ABC):
--- a/blurr/core/base.py
+++ b/blurr/core/base.py
@@ -1,27 +1,10 @@
 from typing import Dict, Any, Type
 
-import yaml
 from abc import ABC, abstractmethod
-from yamale.schema import Schema, Data
-from yamale.validators import Validator, DefaultValidators
 
-from blurr.core.errors import InvalidSchemaError, SnapshotError
+from blurr.core.errors import SnapshotError
 from blurr.core.evaluation import Expression, EvaluationContext
-from blurr.core.loader import TypeLoader, SCHEMA_MAP
-
-
-# This would be in a more central place along with other validation code
-class DTCType(Validator):
-    TAG = 'dtc_type'
-
-    # all valid types would go here, I just made this up
-    VALUES = list(SCHEMA_MAP.keys()) + ['ProductML:DTC:Streaming']
-
-    def _is_valid(self, value):
-        return value in self.VALUES
-
-    def get_name(self):
-        return "DTC Valid Type"
+from blurr.core.loader import TypeLoader
 
 
 class BaseSchema(ABC):
@@ -35,26 +18,11 @@ class BaseSchema(ABC):
     ATTRIBUTE_TYPE = 'Type'
     ATTRIBUTE_WHEN = 'When'
 
-    SCHEMA = '''
-        Name: str(min=1)
-        Type: dtc_type()
-        When: str(min=1, required=False)
-    '''
-
     def __init__(self, spec: Dict[str, Any]) -> None:
         """
         A schema object must be initialized with a schema spec
         :param spec: Dictionary representation of the YAML schema spec
-        """
-        validators = DefaultValidators.copy()
-        validators['dtc_type'] = DTCType
-
-        self.schema = Schema(
-            yaml.load(self.SCHEMA),
-            name=self.__class__.__name__,
-            validators=validators)
-
-        self.__validate_spec(spec)
+       sh """
         self.__load_spec(spec)
 
     @abstractmethod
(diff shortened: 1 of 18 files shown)


blurr/core/base.py
```diff
@@ -55,50 +53,6 @@ def __load_spec(self, spec: Dict[str, Any]) -> None:
...
-                         '\n\tAttribute: {attribute}'
-                         '\n\tError Message: {message}') \
-            .format(
-            name=spec.get(self.ATTRIBUTE_NAME, str(spec)),
-            attribute=attribute,
-            message=message)
-        raise InvalidSchemaError(error_message)
-
 
 class BaseSchemaCollection(BaseSchema, ABC):
```
prashastk: Is the validate method in this class still needed?

blurr/core/syntax/dtc_streaming_schema.yml
```diff
@@ -0,0 +1,47 @@
...
+Identity: expression()
+Time: expression(required=False)
+When: expression(required=False)
+Store: include('dynamodb_store')
+DataGroups: list(any(include('data_group_variable'), include('data_group_identity'), include('data_group_session')), min=1)
+
+---
+
+data_group_variable:
+  Name: identifier()
```
prashastk: Is there a way to specify that all data group names be unique?

blurr/core/syntax/schema_validator.py
```diff
@@ -0,0 +1,91 @@
+import ast
+
```
prashastk: unnecessary newline

blurr/core/syntax/schema_validator.py
```diff
@@ -0,0 +1,91 @@
...
+import re
+from yamale import yamale
+from yamale.schema import Data
+from yamale.validators import DefaultValidators, Validator
+
+from blurr.core.errors import InvalidSchemaError
+
+
+def is_expression(s) -> bool:
```
prashastk: Typing for s?

prashastk: Also applicable to other functions here.

blurr/core/syntax/schema_validator.py
```diff
@@ -0,0 +1,91 @@
...
+    def _is_valid(self, value):
+        return is_expression(value)
+
+    def get_name(self):
+        return "Expression"
+
+
+VALIDATORS = {
+    **DefaultValidators.copy(), "data_type": DataType,
+    "identifier": Identifier,
```
prashastk: Lets use Identifier.TAG here instead of re-defining the constant. Same for expression below.

blurr/core/syntax/schema_validator.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def _is_valid(self, value):
+        return is_expression(value)
+
+    def get_name(self):
+        return "Expression"
+
+
+VALIDATORS = {
+    **DefaultValidators.copy(), "data_type": DataType,
```
prashastk: Use single quote (') for strings. That is what we have been using everywhere else.

Note that for docstring we use double quotes (") as """

blurr/runner/local_runner.py
```diff
@@ -1,10 +1,13 @@
...
+import yaml
+from dateutil import parser
+
+from blurr.core.evaluation import Context
 from blurr.core.record import Record
-from blurr.core.evaluation import Context, EvaluationContext
+from blurr.core.streaming_transformer import StreamingTransformerSchema, StreamingTransformer
+from blurr.core.syntax.schema_validator import validate
 from blurr.store.local_store import LocalStore
-from dateutil import parser
```
prashastk: Thanks for sorting this out. I wonder if yapf can handle standardization of imports.

blurr/core/base.py
mserranom: Not for syntax validation, but I didn't have enough info about whether there's other type of validation expected

blurr/core/syntax/dtc_streaming_schema.yml
mserranom: That won't belong in syntax validation. Think on a compiler, whether a class name is duplicated is a concern during semantic analysis, a grammar has no context of such thing.

blurr/core/syntax/schema_validator.py
mserranom: will add typing.

Created this https://github.com/productml/blurr/issues/35

blurr/core/syntax/schema_validator.py
mserranom: Will do. I also expected yapf to take care of this.

blurr/core/syntax/schema_validator.py
mserranom: Ah yes. I had a problem with that, I left as it is with the idea to go back and reuse the constant, but seems I forgot. Will update this.

blurr/runner/local_runner.py
mserranom: I think I just used IntelliJ's "Optimise imports"

PR #27, opened 2018-03-16
Storage implementation

indradeep:
Storage interface is available that is able to do gets by data group, and by date range.  Unit tests are included.

Session storage is now working and validated.


tests/runner/data/sample.yml
```diff
@@ -7,20 +7,18 @@ Name: Sessions
 Identity: source.user_id
 Time: parser.parse(source.event_time)
 
-Store:
-   - Type: ProductML:DTC:Storage:DynamoDB
-     Name: dynamo
-     Retention: 180
-     Table: ProductML_Test_Sessions
+Stores:
```
mserranom: Update the syntax files to support this store, otherwise tests will break

blurr/core/store.py
```diff
@@ -1,44 +1,159 @@
...
 
-        self._cache: Dict[str, Any] = dict()
+    def __init__(self, identity: str, group: str, timestamp: datetime = None) -> None:
+        """
+        Initializes a new key for storing data
+        :param identity: Primary identity of the record being stored
+        :param group: Secondary identity of the record
+        :param timestamp: Optional timestamp that can be used for time range queries
+        """
+        if identity.isspace() or group.isspace():
```
prashastk: isspace() is true only for ' '. Its not true for ''. I don't think that is what you are going for here?

blurr/core/store.py
```diff
@@ -1,44 +1,159 @@
...
+            raise ValueError('`identity` and `value` must be present.')
 
-    def _cache_get(self, identity: str, group: str) -> Any:
-        return self._cache.get(identity, {}).get(group, None)
+        self.identity = identity
+        self.group = group
+        self.timestamp = timestamp if not timestamp or timestamp.tzinfo else timestamp.replace(tzinfo=timezone.utc)
```
prashastk: How does this work when timestamp is None?

blurr/core/loader.py
```diff
@@ -37,6 +41,10 @@ def load_schema(type_name: str):
     def load_item(type_name: str):
         return TypeLoader.load_type(type_name, ITEM_MAP)
 
+    @staticmethod
+    def load_store(type_name: str):
```
prashastk: Where is this being used?

tests/runner/data/sample.yml
```diff
@@ -17,10 +17,10 @@ DataGroups:
 
  - Type: ProductML:DTC:DataGroup:SessionAggregate
    Name: session
-   Store: dynamo
+   Store: memory
```
prashastk: Putting the store in here is very confusing when the same dtc can potentially work using a memroy store in the offline ML dev scenario and dynamo in a production scenario.

We should think about how this can be simplified.

blurr/runner/local_runner.py
```diff
@@ -45,6 +48,11 @@ def execute(self):
         for file in self._raw_files:
             self._execute_file(file)
 
+        # Ensure that the final values are saved
+        for item in self._user_transformer.values():
+            item.finalize()
+
+
```
prashastk: Unnecessary newlines?

blurr/core/store.py
```diff
@@ -1,44 +1,159 @@
...
-            item = self._store_get(identity, group)
-            self._cache_save(item)
+            # If not, load from store and put item in cache
+            item = self._store_get(key)
+            if item:
+                self._cache[key] = item
 
         return item
 
+    def get_range(self, start: Key, end: Key = None, count: int = 0) -> Dict[Key, Any]:
```
prashastk: I thought this was going to return a list of SessionDataGroup objects?

blurr/store/memory_store.py
```diff
@@ -0,0 +1,43 @@
...
+    def load(self, state: Dict[str, Any]) -> Any:
+        for k, v in state.items():
+            key = Key.parse(k)
+            self._cache[key] = v
+
+    def _store_get_range(self, start: Key, end: Key = None, count: int = 0) -> Dict[Key, Any]:
+        if not count:
+            return {key: item for key, item in self._cache.items() if start < key < end}
+        else:
+            filter_condition = (lambda i: i[0] > start) if count > 0 else (lambda i: i[0] < start)
```
prashastk: Did you look into [SortedDict](http://www.grantjenks.com/docs/sortedcontainers/sorteddict.html)?

d.index and d.irange will be much more efficient ways of doing this.

blurr/core/store.py
indradeep: `if not timestamp` is true when timestamp is `None`.  the `self.timestamp` will be set to `None`

blurr/core/loader.py
indradeep: Eventually...

tests/runner/data/sample.yml
indradeep: yes... I think it's a good idea to take it out till we can figure out further

blurr/core/store.py
indradeep: No, we cannot make the store aware of the object this came from... we need to have this pass through the DTC somehow.

blurr/store/memory_store.py
indradeep: Good idea. I should take a look.  Added this to the Project tasks

blurr/core/store.py
indradeep: Agreed.  Fixing.

PR #32, opened 2018-03-16
Window DTC implementation

prashastk:
Implemented Window DTC which includes objects and their schema:
- Window Transformer 
- AnchorDataGroup
- Anchor
- Window

Also implemented a SchemaLoader which can be used to load any schema that has been added to the SchemaLoader. This was needed because window would reference a SessionDataGroup and its schema needs to be accessible when the window is being generated.
With the addition of SchemaLoader, the concept of a fully qualified name has also been introduced. E.g. A field in a SesssionAggregate DataGroup would have a fully qualified name of <Stream DTC Name>.<Session Name>.<Field Name>.


blurr/core/anchor.py
```diff
@@ -0,0 +1,62 @@
...
+        self.condition_met: Dict[datetime, int] = {}
+        self.anchor_session = None
+
+    def evaluate_anchor(self, session) -> bool:
+        self.anchor_session = session
+        if self.max_condition_met(session):
+            return False
+
+        if self.evaluate():
+            self.condition_met[self.anchor_session.start_time.date(
```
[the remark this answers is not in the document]
prashastk: Done.

blurr/core/anchor_data_group.py
```diff
@@ -0,0 +1,70 @@
+from typing import Dict, Any
+
+from blurr.core.data_group import DataGroup, DataGroupSchema
+from blurr.core.evaluation import EvaluationContext
+from blurr.core.window import Window
+
+
+class AnchorDataGroupSchema(DataGroupSchema):
```
mserranom: Is the plan for the future to add tests for these classes or just test `window_transformer`? 

This and `anchor.py` have enough logic inside to benefit from an specific unit test.

blurr/core/anchor.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class Anchor(BaseItem):
+    def __init__(self, schema: AnchorSchema,
+                 evaluation_context: EvaluationContext):
+        super().__init__(schema, evaluation_context)
+        self.condition_met: Dict[datetime, int] = defaultdict(int)
+        self.anchor_session = None
+
+    def evaluate_anchor(self, session) -> bool:
```
mserranom: session typings

blurr/core/anchor.py
```diff
@@ -0,0 +1,58 @@
...
+            self.condition_met[self.anchor_session.start_time.date()] += 1
+            return True
+
+        return False
+
+    def evaluate(self) -> bool:
+        return self.schema.condition.evaluate(self.evaluation_context)
+
+    def max_condition_met(self, session) -> bool:
+        if self.condition_met.get(session.start_time.date(),
```
mserranom: wouldn't it be the same:
```python
return self.condition_met.get(session.start_time.date(), 0) >= self.schema.max
```
?

blurr/core/anchor_data_group.py
```diff
@@ -0,0 +1,70 @@
...
+    Manages the generation of AnchorAggregate as defined in the schema.
+    """
+
+    def __init__(self, schema: AnchorDataGroupSchema,
+                 evaluation_context: EvaluationContext) -> None:
+        super().__init__(schema, evaluation_context)
+        self.window = Window(self.schema.window_schema
+                             ) if schema.window_schema is not None else None
+
+    def prepare_window(self, store, identity, start_time):
```
mserranom: typings, here and some places around. We should address https://github.com/productml/blurr/issues/35 soon, the longer it takes the more complicated it'll get

blurr/core/window_transformer.py
```diff
@@ -0,0 +1,82 @@
+from datetime import datetime
+from typing import Any, Dict
+
+from blurr.core.anchor import AnchorSchema, Anchor
+from blurr.core.anchor_data_group import AnchorDataGroup
+from blurr.core.errors import AnchorSessionNotDefinedError
+from blurr.core.evaluation import Context, EvaluationContext
+from blurr.core.schema_loader import SchemaLoader
```
mserranom: not used

blurr/core/anchor_data_group.py
prashastk: Yes, we will be adding those unit tests in a subsequent PR (soon!).

blurr/core/anchor.py
prashastk: Done.

blurr/core/anchor.py
prashastk: Correct. Changed. Also, condition_met is of type defaultdict(int) so we don't need to use `get(key, default)`.

blurr/core/anchor_data_group.py
prashastk: Added. MyPy had given me issues when I briefly tried it out for our old etl library. But, I agree, would be good to have that set up soon so that we can catch these in an automated fashion.

blurr/core/window_transformer.py
prashastk: Removed.

blurr/core/base.py
```diff
@@ -92,16 +99,18 @@ def validate(self, spec: Dict[str, Any]):
...
         Overrides base load to include loads for nested items
         """
 
         # Load nested schema items
         self.nested_schema: Dict[str, Type[BaseSchema]] = {
-            schema_spec[self.ATTRIBUTE_NAME]: TypeLoader.load_schema(
-                schema_spec[self.ATTRIBUTE_TYPE])(schema_spec)
-            for schema_spec in spec[self._nested_item_attribute]
+            schema_spec[self.ATTRIBUTE_NAME]:
+            self.schema_loader.get_schema_object(
```
indradeep: Can we make a utility function as part of the schema loader to get nested schema spec - given the parent and child?

blurr/core/window_transformer.py
```diff
@@ -0,0 +1,80 @@
...
+        return False
+
+    def evaluate(self):
+        if 'anchor' not in self.evaluation_context.global_context or self.anchor.anchor_session is None:
+            raise AnchorSessionNotDefinedError()
+
+        if not self.needs_evaluation:
+            return
+
+        for _, item in self.nested_items.items():
```
indradeep: Consider using `for item in self.nested_items.values():`

blurr/core/store.py
```diff
@@ -1,6 +1,6 @@
 from abc import ABC, abstractmethod
```
indradeep: This will need some refactoring when merging with the latest store changes.

prashastk: Ack.

blurr/core/window_transformer.py
prashastk: Done.

blurr/core/base.py
prashastk: Done.

PR #33, opened 2018-03-17
Added Variable and Identity data groups

indradeep:
Created two new data groups as per spec.  I'm still dissatisfied with the levels of awareness that the Transformer needs to have about the data groups.... wish there was a better way.

### DIFF
--- a/blurr/core/record.py
+++ b/blurr/core/record.py
@@ -1,9 +1,8 @@
 class Record:
     def __init__(self, event_dict):
-        self._event_dict = event_dict
+        for k, v in event_dict.items():
+            self.__dict__[k] = v
 
-    def __getattr__(self, item):
-        if isinstance(self._event_dict[item], dict):
-            return Record(self._event_dict[item])
-        else:
-            return self._event_dict[item]
+    def __getattr__(self, name):
+        if not hasattr(super(), name):
+            return None
--- a/blurr/core/syntax/dtc_streaming_schema.yml
+++ b/blurr/core/syntax/dtc_streaming_schema.yml
@@ -12,7 +12,7 @@ DataGroups: list(any(include('data_group_variable'), include('data_group_identit
 
 data_group_variable:
   Name: identifier()
-  Type: enum('ProductML:DTC:DataGroup:Variable')
+  Type: enum('ProductML:DTC:DataGroup:VariableAggregate')
   Store: identifier(required=False)
   When: expression(required=False)
   Fields: list(include('field'), min=1)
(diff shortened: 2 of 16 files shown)


blurr/core/record.py
```diff
@@ -1,9 +1,8 @@
 class Record:
     def __init__(self, event_dict):
-        self._event_dict = event_dict
+        for k, v in event_dict.items():
+            self.__dict__[k] = v
 
-    def __getattr__(self, item):
-        if isinstance(self._event_dict[item], dict):
```
prashastk: Removed dict in a dict support?

blurr/core/syntax/dtc_streaming_schema.yml
```diff
@@ -12,7 +12,7 @@ DataGroups: list(any(include('data_group_variable'), include('data_group_identit
 
 data_group_variable:
   Name: identifier()
-  Type: enum('ProductML:DTC:DataGroup:Variable')
+  Type: enum('ProductML:DTC:DataGroup:VariableAggregate')
```
prashastk: Should we just call it Variable?

blurr/core/transformer.py
```diff
@@ -51,4 +52,5 @@ def __init__(self, store: Store, schema: TransformerSchema, identity: str,
 
     def finalize(self):
         for item in self.nested_items.values():
-            self.store.save(Key(self.identity, item.name), item.snapshot)
+            if not isinstance(item, VariableDataGroup):
+                self.store.save(Key(self.identity, item.name), item.snapshot)
```
prashastk: Maybe add a save function to the DataGroups which takes in a store? VariableDataGroup will do nothing with it.

blurr/core/record.py
indradeep: Yes.  It adds items to it's own object dictionary.

blurr/core/transformer.py
indradeep: This has been addressed in a future PR.

blurr/core/syntax/dtc_streaming_schema.yml
indradeep: I'll leave it as is now... le'ts address naming in a separate round.

blurr/core/record.py
prashastk: How does dict in a dict work then? source.some_dict_in_streaming_data.value?

indradeep: Fixed.  All nested dictionaries will also become records.

PR #36, opened 2018-03-19
Storage design refactoring

indradeep:
Refactored store to be instantiated in the Transformer and be used in the Data Group.

Changed signatures for Data Group to be responsible for storing state.

Removed unnecessary validation

Removed caching from the store implementation


blurr/core/base.py
```diff
@@ -25,14 +25,6 @@ def __init__(self, spec: Dict[str, Any]) -> None:
         """
         self.__load_spec(spec)
 
-    @abstractmethod
```
indradeep: All validation has moved to the Yamle validator

tests/core/syntax/dtcs/invalid_string_instead_integer.yml
```diff
@@ -1,24 +0,0 @@
-Type: ProductML:DTC:Streaming
```
indradeep: This file is no longer relevant - We'll include such checks when we implement Dynamo Store

blurr/core/data_group.py
```diff
@@ -28,7 +32,7 @@ class DataGroup(BaseItemCollection):
     to be called as properties of the data group itself.
     """
 
-    def __init__(self, schema: DataGroupSchema,
+    def __init__(self, schema: DataGroupSchema, identity: Any,
```
prashastk: Should the identity be of type str?

blurr/core/session_data_group.py
```diff
@@ -93,5 +76,17 @@ def split_now(self) -> bool:
 
         return False
 
-    def reset(self):
-        self.__init__(self.schema, self.evaluation_context)
+    def evaluate(self) -> None:
```
mserranom: this is all untested. It won't be merged into core until reasonable coverage is added right?

blurr/core/session_data_group.py
```diff
@@ -93,5 +76,17 @@ def split_now(self) -> bool:
...
+        """
+        Evaluates the current item
+        """
+
+        # If a split is imminent, save the current session snapshot with the timestamp
+        if not (self.schema.split is None or self.start_time is None or self.end_time is None) and \
```
mserranom: long boolean expressions are more readable if extracted to variables first as in:

```
split_is_imminent = not(self.schema.split is None or self.start_time is None or self.end_time is None)
whatever_this_means = self.schema.split.evaluate(self.evaluation_context)
if split_is_imminent and whatever_this_means:
 ...
```

blurr/core/syntax/dtc_streaming_schema.yml
```diff
@@ -39,9 +38,6 @@ field:
   When: expression(required=False)
   Atomic: bool(required=False)
 
-dynamodb_store:
```
mserranom: +1 for this: if we're implementing dynamodb yet, better removing than leaving there for an eventual implementation

blurr/core/syntax/dtc_streaming_schema.yml
```diff
@@ -5,15 +5,14 @@ Name: identifier()
 Identity: expression()
 Time: expression(required=False)
 When: expression(required=False)
-Store: include('dynamodb_store')
+Stores: List(any(include('memory_store')))
```
mserranom: I guess it should be `Stores: List(any(include('memory_store')), min=1)`

blurr/core/transformer.py
```diff
@@ -15,42 +16,61 @@ class TransformerSchema(BaseSchemaCollection, ABC):
...
-
     def load(self, spec: Dict[str, Any]) -> None:
-        # Ensure that the base loader is invoked
-        super().load(spec)
-
         # Load the schema specific attributes
         self.version = spec[self.ATTRIBUTE_VERSION]
         self.description = spec[self.ATTRIBUTE_DESCRIPTION]
 
+        # Load list of stores from the schema
```
mserranom: Would it make sense to extract this type of code to a wrapper of the DTC validator?

In my head it would be the equivalent of a AST class, using the DTC schemas as grammar, something like:

```python
class DTCStreamingAbstractSyntaxTree:
    def __init__(self, dtc_yam_file:strl)
        pass ...
        self.validate() # perform validation
      
    def get_stores(self) -> dict:
       return {
              schema_spec[self.ATTRIBUTE_NAME]: TypeLoader.load_store(
                     schema_spec[self.ATTRIBUTE_TYPE])(schema_spec)
                     or schema_spec in spec[self.ATTRIBUTE_STORES]
               }
```

blurr/store/memory_store.py
```diff
@@ -4,18 +4,26 @@
...
-            key = Key.parse(k)
-            self._cache[key] = v
-
-    def _store_get_range(self, start: Key, end: Key = None,
-                         count: int = 0) -> Dict[Key, Any]:
+    """
+    In-memory store implementation
+    """
+
+    def __init__(self, spec: Dict[str, Any] = None) -> None:
```
mserranom: why optional `spec`? That's not covered in the test

blurr/store/memory_store.py
```diff
@@ -4,18 +4,26 @@
 
 
 class MemoryStore(Store):
```
mserranom: I understand you're delaying some tests, but this looks like should be tested at this point. 

How can we make sure that we don't forget anything later? I fear not that tests will be missing, but a lack of thoroughness. Personally I'm only able to think on all the required test cases when I work in a single unit.

tests/core/syntax/schema_validator_test.py
```diff
@@ -52,14 +52,6 @@ def test_invalid_wrong_version():
         validate(dtc_dict)
     assert "Version: '2088-03-01' not in ('2018-03-01',)" in str(err.value)
 
-
-def test_invalid_string_instead_of_integer():
```
mserranom: ` We'll include such checks when we implement Dynamo Store`

I don't understand why this is deleted? The purpose of this test is to make sure datatype error reporting is correct?

blurr/core/session_data_group.py
vigbk: or make it into a function

blurr/core/session_data_group.py
indradeep: The unit tests will be added before merging to master

blurr/core/transformer.py
indradeep: I'm not sure I follow.  I do not see how these components are part of the abstract syntax tree.  The reason why the store is part of the schema is because the same store is used for all user instances for a particular data groups

blurr/store/memory_store.py
indradeep: Good point...this is just for moving fast.  we really need to get to this. I agree.

blurr/core/transformer.py
mserranom: ok, I was talking about a different, but related thing. Let's leave it here, and I'll try to phrase a proposal.

tests/core/syntax/schema_validator_test.py
indradeep: I've added it back.  The test was tied to Dynamo which i have removed.  I now tied this test case to the Window DTC and Anchor Max field.

PR #37, opened 2018-03-20
changes after PR 25 Feedback

mserranom:
PR https://github.com/productml/blurr/pull/25

### DIFF
--- a/blurr/core/syntax/schema_validator.py
+++ b/blurr/core/syntax/schema_validator.py
@@ -1,4 +1,5 @@
 import ast
+from typing import Dict
 
 import re
 from yamale import yamale
@@ -48,7 +49,7 @@ class Identifier(Validator):
 class Expression(Validator):
     TAG = 'expression'
 
-    def _is_valid(self, value) -> bool:
+    def _is_valid(self, value: str) -> bool:
         return is_expression(value)
 
     def get_name(self) -> str:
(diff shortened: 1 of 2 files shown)


blurr/core/syntax/schema_validator.py
```diff
@@ -68,24 +68,24 @@ def get_name(self):
     'blurr/core/syntax/dtc_window_schema.yml', validators=VALIDATORS)
 
 
-def _validate_window(dtc_dict, name):
+def _validate_window(dtc_dict: dict, name: str) -> None:
```
vigbk: The type is Dict. https://docs.python.org/3/library/typing.html

blurr/core/syntax/schema_validator.py
```diff
@@ -28,37 +28,37 @@ class DataType(Validator):
...
-        return "Identifier"
+    def get_name(self) -> str:
+        return 'Identifier'
 
 
 class Expression(Validator):
     TAG = 'expression'
 
-    def _is_valid(self, value):
+    def _is_valid(self, value) -> bool:
```
prashastk: value should be str?

blurr/core/syntax/schema_validator.py
mserranom: fixed

blurr/core/syntax/schema_validator.py
mserranom: fixed

blurr/core/syntax/schema_validator.py
mserranom: fixed

PR #42, opened 2018-03-20
Remove BaseSchema load methods and do the loading as part of init.

prashastk:
Remove BaseSchema load methods and do the loading as part of init.

This makes the loading more transparent and all instance variables are defined clearly in one place across the layers of abstraction.

### DIFF
--- a/blurr/core/base.py
+++ b/blurr/core/base.py
@@ -27,29 +27,19 @@ class BaseSchema(ABC):
         """
         self.schema_loader = schema_loader
         self.fully_qualified_name = fully_qualified_name
-        self.__load_spec()
-
-    @abstractmethod
-    def load(self) -> None:
-        """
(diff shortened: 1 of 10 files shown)


blurr/core/base.py
```diff
@@ -27,29 +27,19 @@ def __init__(self, fully_qualified_name: str,
...
+
         self.name: str = self._spec[self.ATTRIBUTE_NAME]
         self.type: str = self._spec[self.ATTRIBUTE_TYPE]
         self.when: Expression = Expression(
             self._spec[self.ATTRIBUTE_WHEN]
         ) if self.ATTRIBUTE_WHEN in self._spec else None
 
-        # Invokes the loads of the subclass
-        self.load()
+    def extend_schema(self):
```
mserranom: This is an abstract method? Some comment would be helpful

PR #43, opened 2018-03-21
field-unit-test  Added unit tests for Field

vigbk:


tests/core/field_test.py
```diff
@@ -0,0 +1,94 @@
...
+from blurr.core.schema_loader import SchemaLoader
+
+@fixture
+def test_field_schema() -> Dict[str, Any]:
+    schema_loader = SchemaLoader()
+    name = schema_loader.add_schema(yaml.load('''
+Name: max_attempts
+Type: integer
+Value: 5
+'''))
```
prashastk: instead of yaml load should we pass in a dict?

tests/core/field_test.py
```diff
@@ -0,0 +1,94 @@
...
+    assert field.snapshot == 0
+    
+    field.evaluate()
+    assert field.snapshot == 5
+
+
+def test_field_restore(test_field_schema):
+    field = MockField(test_field_schema, EvaluationContext())
+    field.restore(5)
+    assert field.value == 5
```
prashastk: newline at end

tests/core/field_test.py
vigbk: sure. will change that.

tests/core/field_test.py
vigbk: Noted

PR #49, opened 2018-03-22
Tests for window and anchor

prashastk:
Tests for window and anchor
Moved some common test fixtures to conftest so that they can be re-used."

### DIFF
--- a/blurr/core/anchor.py
+++ b/blurr/core/anchor.py
@@ -4,6 +4,7 @@ from typing import Dict, Any
 
 from blurr.core.base import BaseSchema, BaseItem
 from blurr.core.evaluation import Expression, EvaluationContext
+from blurr.core.schema_loader import SchemaLoader
 from blurr.core.session_data_group import SessionDataGroup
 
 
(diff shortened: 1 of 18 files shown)


prashastk:
@mserranom can you be more specific about what other fixture should be tested? these tests are trying to test the business logic in the window and anchor classes / files.

mserranom:
`can you be more specific about what other fixture should be tested?`

This is a tricky question. I'm not sure what's a good fixture set to test because I didn't participate in writing the code, therefore I wouldn't know the specification. At personal level, I design the fixtures while writing the code to make sure the code does what I do.

More than "let's increase the coverage", my question had 2 parts:
  1. Are you guys happy with the tests? You wrote the code, you should have a degree of confidence with the spec. It just feels incomplete to me, I won't be able to say if it's worth worrying.
  2.  If we decide we should probably need more tests, what should we do? Thinking out loud, possibly a topdown strategy might work here, adding a lot of e2e tests, and add coverage if issues emerge.

prashastk:
I am bit confused about whether your comment is about this PR or our current state of unit tests as whole.

I agree with you that the coverage as a whole is not there yet (and we are still working on it!). Having more complex E2E tests is a good approach for a system like this where various classes and levels of abstraction need to work with each other to do something meaningful.

PR #50, opened 2018-03-23
Data group unit tests

vigbk:


blurr/core/data_group.py
```diff
@@ -27,12 +28,19 @@ def __init__(self, fully_qualified_name: str,
...
                          self.ATTRIBUTE_FIELDS)
         self.store = None
+
         if self.ATTRIBUTE_STORE in self._spec:
             store_fq_name = self.schema_loader.get_fully_qualified_name(
                 self.schema_loader.get_transformer_name(
                     self.fully_qualified_name),
                 self._spec[self.ATTRIBUTE_STORE])
-            self.store = self.schema_loader.get_schema_object(store_fq_name)
+            try:
```
indradeep: This exception is best raised from inside the Schema loader.  The exception should bubble up at this stage.

tests/core/data_group_schema_test.py
```diff
@@ -0,0 +1,49 @@
...
+from pytest import fixture
+import pytest
+
+from blurr.core.data_group import DataGroupSchema
+from blurr.core.errors import InvalidSchemaError
+from blurr.core.schema_loader import SchemaLoader
+from blurr.store.memory_store import MemoryStore
+
+
+def get_data_group_schema_spec() -> Dict[str, Any]:
```
indradeep: Why not declare these gets as as fixtures?

tests/core/data_group_test.py
```diff
@@ -0,0 +1,102 @@
...
+        evaluation_context=EvaluationContext())
+    nested_items = data_group.nested_items
+    assert len(nested_items) == 1
+    assert "event_count" in nested_items
+    assert isinstance(nested_items["event_count"], Field)
+
+
+def test_data_group_persist(data_group_schema_with_store,
+                            data_group_schema_without_store):
+    data_group = DataGroup(
```
indradeep: DataGroup should be an ABC.  Can you make that change? 

That will also influence your test cases.

tests/core/data_group_test.py
```diff
@@ -0,0 +1,102 @@
...
+        schema=data_group_schema_with_store,
+        identity="12345",
+        evaluation_context=EvaluationContext())
+    dt = datetime.now()
+    dt.replace(tzinfo=timezone.utc)
+    data_group.persist(dt)
+    snapshot_data_group = data_group.schema.store.get(
+        Key(identity="12345", group="user", timestamp=dt))
+    assert snapshot_data_group is not None
+    assert snapshot_data_group == data_group.snapshot
```
indradeep: I don't see the purpose of the previous `data_group.persist()` on line 78.  The object that is referred to in the assert  is the new data group object.

blurr/core/data_group.py
vigbk: Sure

tests/core/data_group_schema_test.py
vigbk: Will change it

tests/core/data_group_test.py
vigbk: Will do

tests/core/data_group_test.py
vigbk: that is to make sure that there are no errors when you have a datagroup without a store and call persist.

blurr/core/data_group.py
mserranom: would be nice to extract this to a well-named function for clarity and maintainability.

tests/core/data_group_schema_test.py
```diff
@@ -0,0 +1,49 @@
...
+
+def get_store_spec() -> Dict[str, Any]:
+    return {'Type': 'ProductML:DTC:Store:MemoryStore', 'Name': 'memory'}
+
+
+def test_data_group_schema_initialization_with_store():
+    schema_loader = SchemaLoader()
+    name = schema_loader.add_schema(get_data_group_schema_spec())
+    with pytest.raises(
+            InvalidSchemaError, match="Store memory not declared is schema"):
```
mserranom: nice! wasn't aware of this `match`

blurr/core/data_group.py
vigbk: the exception or loading the store ?

tests/core/data_group_schema_test.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def get_store_spec() -> Dict[str, Any]:
+    return {'Type': 'ProductML:DTC:Store:MemoryStore', 'Name': 'memory'}
+
+
+def test_data_group_schema_initialization_with_store():
+    schema_loader = SchemaLoader()
+    name = schema_loader.add_schema(get_data_group_schema_spec())
+    with pytest.raises(
```
mserranom: This are 2 different test cases, the one throwing the InvalidSchemaError and the happy path with the store spec.

tests/core/data_group_schema_test.py
```diff
@@ -0,0 +1,49 @@
...
+            InvalidSchemaError, match="Store memory not declared is schema"):
+        data_group_schema = DataGroupSchema(name, schema_loader)
+
+    schema_loader.add_schema(get_store_spec(), 'user')
+    data_group_schema = DataGroupSchema(name, schema_loader)
+    assert data_group_schema.store is not None
+    assert data_group_schema.store.name == 'memory'
+
+
+def test_data_group_schema_initialization_without_store():
```
mserranom: A more general comment / side note. A glance at the test title should be enough to understand the spec.

A rule of thumb I've used successfully in the past is that test names should always have at least 1 verb. For this case, I'd have used something like `test_data_group_schema_can_be_initialized_without_store`

Unfortunately this is something I keep seeing in python projects all around github. Maybe cryptic names is a python idiom :/

Any thoughts/ideas?

blurr/core/schema_loader.py
```diff
@@ -92,7 +93,11 @@ def get_schema_spec(self, fully_qualified_name: str) -> Dict[str, Any]:
         :param fully_qualified_name: The fully qualified name of the schema needed.
         :return: Schema dictionary.
         """
-        return self._spec[fully_qualified_name]
+        try:
+            return self._spec[fully_qualified_name]
+        except:
+            raise InvalidSchemaError(
+                "{} not declared is schema".format(fully_qualified_name))
```
prashastk: not declared 'in' schema?

tests/core/data_group_schema_test.py
```diff
@@ -0,0 +1,56 @@
...
+from blurr.store.memory_store import MemoryStore
+
+
+@fixture
+def data_group_schema_spec() -> Dict[str, Any]:
+    return {
+        'Type': 'ProductML:DTC:DataGroup:SessionAggregate',
+        'Name': 'user',
+        'Store': 'memory',
+        'Filter': 'source.event_id in ["app_launched", "user_updated"]',
```
prashastk: Filter should be When. Or remove it since its not being used?

tests/core/data_group_test.py
```diff
@@ -0,0 +1,110 @@
...
+from blurr.core.store import Key
+from blurr.store.memory_store import MemoryStore
+
+
+def get_data_group_schema_spec() -> Dict[str, Any]:
+    return {
+        'Type': 'ProductML:DTC:DataGroup:SessionAggregate',
+        'Name': 'user',
+        'Store': 'memory',
+        'Filter': 'source.event_id in ["app_launched", "user_updated"]',
```
prashastk: This should be When. Also, can you test evaluate?

tests/core/data_group_test.py
prashastk: Lets break it into two tests then? test_data_group_persist_no_store and test_data_group_persist_with_store.

tests/core/data_group_test.py
vigbk: Does the evaluate test belongs here ? The DataGroup class does nothing special with evaluate ?

tests/core/data_group_schema_test.py
vigbk: sure

tests/core/data_group_test.py
vigbk: sure.

blurr/core/schema_loader.py
vigbk: Noted

tests/core/data_group_test.py
prashastk: Hmmmm BaseItemCollection implements those. Looks like we don't have any tests for that. Can you raise an issue for that?

PR #56, opened 2018-03-23
Record unit tests

indradeep:
- Simplified the Record object to support the following:
** Nested Dictionaries as records
** Nested arrays with dictionaries as Record Lists with Records

- Added unit tests

- Cleaned up code

### DIFF
--- a/tests/core/record_test.py
+++ b/tests/core/record_test.py
@@ -1,0 +1,56 @@
+from pytest import fixture
+
+from blurr.core.record import Record
+
+
+def test_simple_fields() -> None:
+    record = Record({
+        'string_field': 'string value',
+        'int_field': 1
+    })
+
+    assert record.string_field == 'string value'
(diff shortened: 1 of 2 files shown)


tests/core/record_test.py
```diff
@@ -0,0 +1,56 @@
...
+                'field_1': 'one',
+                'field_2': 2
+            },
+            ['one', 'two'],
+            1
+        ]
+    })
+    assert len(record.complex_array_field) == 3
+    assert record.complex_array_field[0].field_1 == 'one'
+    assert record.complex_array_field[1][0] == 'one'
```
vigbk: why not just compare it to ['one', 'two']

indradeep: To ensure that the access patterns are maintained - dots, brackets, etc. which will not be in the checking path if object comparison is done.

PR #67, opened 2018-03-26
Added support for window dtc to the local runner

### LINKED ISSUE #65: Local runner (with window support)

prashastk:
Fixes #65 

The local runner code has been split into an identity_runner as well. identity_runner util will also be used when DTC execution is being done in spark.

Decided not to use Trie as we are now grouping data by identity and running operations on it.

Also, added missing Key class unit tests.

### DIFF
--- a/blurr/core/base.py
+++ b/blurr/core/base.py
@@ -1,4 +1,4 @@
-from typing import Dict, Any, Type
+from typing import Dict, Any, Type, TypeVar
 
 from abc import ABC, abstractmethod
 
@@ -66,6 +66,9 @@ class BaseSchemaCollection(BaseSchema, ABC):
         }
 
 
+BaseItemT = TypeVar('T', bound='BaseItem')
+
+
 class BaseItem(ABC):
     """
     Base class for for all leaf items that do not contain sub-items
@@ -116,7 +119,7 @@ class BaseItem(ABC):
         raise NotImplementedError('snapshot() must be implemented')
 
     @abstractmethod
-    def restore(self, snapshot) -> 'BaseItem':
+    def restore(self, snapshot) -> BaseItemT:
         """
         Restores the state of an item from a snapshot
         """
@@ -164,7 +167,7 @@ class BaseItemCollection(BaseItem):
             print('Error while creating snapshot for {}', self.name)
             raise SnapshotError(e)
 
-    def restore(self, snapshot: Dict[str, Any]) -> BaseItem:
+    def restore(self, snapshot: Dict[str, Any]) -> 'BaseItemCollection':
         """
         Restores the state of a collection from a snapshot
         """
--- a/blurr/core/base.py
+++ b/blurr/core/base.py
@@ -4,7 +4,6 @@ from abc import ABC, abstractmethod
 
 from blurr.core.errors import SnapshotError
 from blurr.core.evaluation import Expression, EvaluationContext
-from blurr.core.loader import TypeLoader
 from blurr.core.schema_loader import SchemaLoader
 
 
--- a/blurr/core/base.py
+++ b/blurr/core/base.py
@@ -65,7 +65,7 @@ class BaseSchemaCollection(BaseSchema, ABC):
         }
 
 
-BaseItemT = TypeVar('T', bound='BaseItem')
+BaseItemType = TypeVar('T', bound='BaseItem')
 
 
 class BaseItem(ABC):
@@ -118,7 +118,7 @@ class BaseItem(ABC):
         raise NotImplementedError('snapshot() must be implemented')
 
     @abstractmethod
-    def restore(self, snapshot) -> BaseItemT:
+    def restore(self, snapshot) -> BaseItemType:
         """
         Restores the state of an item from a snapshot
         """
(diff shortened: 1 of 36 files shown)


blurr/core/base.py
```diff
@@ -163,7 +166,7 @@ def snapshot(self) -> Dict[str, Any]:
             print('Error while creating snapshot for {}', self.name)
             raise SnapshotError(e)
 
-    def restore(self, snapshot: Dict[str, Any]) -> BaseItem:
+    def restore(self, snapshot: Dict[str, Any]) -> 'BaseItemCollection':
```
vigbk: What do the quotes mean in 'BaseItemCollection'?

blurr/core/window.py
```diff
@@ -58,6 +59,22 @@ def prepare(self, identity: str, start_time: datetime) -> None:
                     Key(identity, self.schema.source.name, start_time), None,
                     self.schema.value), identity)
 
+        self._validate_view()
+
+    def _validate_view(self):
+        if self.schema.type == 'count':
```
mserranom: could be a single statement: `cond1 and cond2`

blurr/core/window_transformer.py
```diff
@@ -51,9 +54,16 @@ def evaluate_anchor(self, session: SessionDataGroup) -> bool:
...
             session.schema.fully_qualified_name, session)
         if self.anchor.evaluate_anchor(session):
-            self.evaluation_context.global_add('anchor', session)
-            self.evaluate()
-            del self.evaluation_context.global_context['anchor']
+
+            try:
+                self.evaluation_context.global_add('anchor', session)
+                self.evaluate()
+                self.anchor.add_condition_met()
```
mserranom: I find much more readable to have the `return True` at the end of the catch than after the `try/catch/finally` block.

May be just me!

tests/runner/local_runner_test.py
```diff
@@ -1,41 +1,56 @@
-from datetime import datetime, timezone
+from datetime import datetime
+
+from dateutil.tz import tzutc
 
 from blurr.core.store import Key
 from blurr.runner.local_runner import LocalRunner
 
 
 def test_local_runner():
```
mserranom: instead of having a big test with multiple assertions would be nicer to have individual tests with meaningful names.

tests/core/window_test.py
```diff
@@ -48,14 +50,22 @@ def test_window_type_day(window: Window) -> None:
...
-    window.prepare('user1', datetime(2018, 3, 7, 21, 36, 31, 0, timezone.utc))
+    with pytest.raises(
+            PrepareWindowMissingSessionsError,
+            match='No matching sessions found'):
+        window.prepare('user1',
+                       datetime(2018, 3, 7, 21, 36, 31, 0, timezone.utc))
     assert window.events == []
 
 
 def test_window_type_hour(window: Window) -> None:
```
mserranom: This test can be broken down into individual, atomic tests, always a good practice.

tests/core/window_test.py
```diff
@@ -48,14 +50,22 @@ def test_window_type_day(window: Window) -> None:
 
     window.schema.type = 'day'
     window.schema.value = 0
-    window.prepare('user1', datetime(2018, 3, 7, 21, 36, 31, 0, timezone.utc))
+    with pytest.raises(
+            PrepareWindowMissingSessionsError,
+            match='No matching sessions found'):
```
mserranom: nice, wasn't aware of this match parameter

blurr/core/base.py
prashastk: Quotes are used for forward reference. This function is in BaseItemCollection itself so 'BaseItemCollection' tells typing that this would be ready in the future.

blurr/core/window.py
prashastk: Done.

blurr/core/window_transformer.py
prashastk: Yea, makes sense. Done.

tests/runner/local_runner_test.py
prashastk: Done.

tests/core/window_test.py
prashastk: Done.

blurr/core/base.py
```diff
@@ -65,6 +65,9 @@ def __init__(self, fully_qualified_name: str, schema_loader: SchemaLoader,
         }
 
 
+BaseItemT = TypeVar('T', bound='BaseItem')
```
indradeep: Can you name this `BaseItemType` so that it's clearer

blurr/core/base.py
indradeep: Why not define a type variable for this too?

blurr/core/store.py
```diff
@@ -51,12 +54,26 @@ def __eq__(self, other: 'Key') -> bool:
                                     other.timestamp)
 
     def __lt__(self, other: 'Key') -> bool:
-        return (self.identity, self.group) == (
-            other.identity, other.group) and self.timestamp < other.timestamp
+        """
```
indradeep: Very interesting.  is this to support the default session - the current session with no timestamp is the latest?

blurr/core/base.py
prashastk: Its not needed. At this point all that typing needs to know that a 'BaseItemCollection' object is being returned.

blurr/core/store.py
prashastk: Yes.

PR #75, opened 2018-03-27
Evaluation Support for operations on Complex Types

indradeep:
# Context
In our DTC, the `Value` of a `Field` is set by taking the return value of the `Value` expression.  This poses a problem for complex objects where operations on these objects may alter the objects themselves, but the object reference is not returned.  For instance, when you evaluate `list.append(value)`, the list itself is altered and nothing is returned.  

One may then consider setting the value of the field directly using the `=` operator.  Explicit sets are not supported by python `eval`.  Setting `=` is supported by `exec`, but that opens a whole host of possibilities on how `Value` expressions are evaluated.  

We needed to make a decision here on whether to support explicit setting of variables from the `Value` expression, and we have come to the conclusion that the complexities introduced by supporting `exec` are not worth the flexibility.  

# Solution
In order to then support operations on complex objects, the developer has two options:
1. Write a function that modifies the object and returns the object as a result
2. We natively overwrite all functions on the base `list`, `set` and `dict` types that return the object itself at the conclusion of the operation.

Solution 1 will be supported as we will offer support for python functions.  However, this is not as clean a way to do fundamental data objects like `list`, `set` and `dict`.  So, it makes sense to implement solution 2 for clarity and usability.  

This PR will implement Solution 2


mserranom:
It sounds to me that what we need are immutable data structures. Instead of having to create or maintain them (optimisation is always tricky) we could use existing solutions such as [pyrsistent](https://github.com/tobgu/pyrsistent).

This would be the example for an [immutable list](https://github.com/tobgu/pyrsistent#pvector).

blurr/core/list_field.py
```diff
@@ -0,0 +1,48 @@
...
+
+from blurr.core.field import FieldSchema
+
+
+class List(list):
+    """
+    Extends native list with operations for evaluation support.
+    """
+
+    def append(self, item):
```
prashastk: Add typing? (here and in all other classes)

blurr/core/list_field.py
```diff
@@ -0,0 +1,48 @@
...
+        """ Removes the specified item"""
+        super().remove(item)
+        return self
+
+    def clear(self):
+        """ Clears all items form the list """
+        super().clear()
+        return self
+
+    def copy(self):
```
prashastk: Do we need to overwrite copy since copy natively returns a list?

blurr/core/map_field.py
```diff
@@ -0,0 +1,43 @@
...
+
+from blurr.core.field import FieldSchema
+
+
+class Map(dict):
+    """
+    Extends native dictionary with operations for evaluation support.
+    """
+
+    def set(self, key, value):
```
prashastk: What about remove / delete ?

blurr/core/list_field.py
```diff
@@ -0,0 +1,48 @@
+from typing import Any
+
+from blurr.core.field import FieldSchema
+
+
+class List(list):
```
prashastk: In a comment mention the rationale behind doing this and explain why other functions are not overridden:
```
def __add__(self, t)
def __setitem__(self, i, v)
```

indradeep:
We do not need these data structures to be mutable.  We need chaining instead.  All operations should return the object itself so that the resulting object can be set to the reference.  

What is implemented is the capability to chain, not immutability.

blurr/core/list_field.py
prashastk: This pattern of:
```
def func(self):
  super().func()
  return self
```

can be done in this manner:
```
class Test(dict):
    def __getattribute__(self, item):
        attr = dict.__getattribute__(self, item)

        def return_self(*args, **kwargs):
            result = attr(*args, **kwargs)
            return self if result is None else result
        return return_self
```

This way each function need not be overridden and chaining support can be added. I'm not sure if there are corner cases here to think about. Would probably need some testing.

indradeep: Removed all explicit method declarations in lieu of the new change.

blurr/core/list_field.py
indradeep: Changed to using native methods

blurr/core/map_field.py
indradeep: Using native implementations.

blurr/core/list_field.py
indradeep: No longer applicable.

blurr/core/list_field.py
indradeep: Implemented.  Please review.

indradeep:
Greatly simplified the solution using Prashast's recommended solution - altering output using function wrappers.  

Also implemented evaluation check so that setting a value using `=` is not allowed.

blurr/core/evaluation.py
```diff
@@ -106,5 +109,15 @@ def evaluate(self, evaluation_context: EvaluationContext) -> Any:
...
-            print('Expression:', self.code_string)
             raise ExpressionEvaluationError(e)
+
+    @staticmethod
+    def validate(code_string: str) -> None:
+        equal_index = code_string.find('=', 0)
+        while equal_index != -1:
+            next_index = equal_index + 1
+            if next_index == len(
+                    code_string) or code_string[next_index] != '=':
```
vigbk: Its possible that a string contains = which would be valid right ? "'Hi my name = Vignesh'". Dont think it would be common though.

blurr/core/evaluation.py
```diff
@@ -106,5 +109,24 @@ def evaluate(self, evaluation_context: EvaluationContext) -> Any:
                         evaluation_context.local_context)
 
         except Exception as e:
-            print('Expression:', self.code_string)
             raise ExpressionEvaluationError(e)
+
+    @staticmethod
+    def validate(code_string: str) -> None:
```
prashastk: ```re.compile('[^!=]+=[^=]+').match(code_string) is not None``` will be a much simpler way to look for = without ==.

blurr/core/field.py
```diff
@@ -100,3 +100,46 @@ def restore(self, snapshot) -> None:
...
+        attribute = super().__getattribute__(item)
+
+        # Return the attribute as-is when it is NOT a function
+        if not callable(attribute):
+            return attribute
+
+        # Wrap the attribute in a function that changes its return value
+        def wrapped_attribute(*args, **kwargs):
+
+            # Execution the underlying method
```
prashastk: Typo: Executing

blurr/core/field.py
```diff
@@ -100,3 +100,46 @@ def restore(self, snapshot) -> None:
...
+            result = attribute(*args, **kwargs)
+
+            # Get the type of the current object
+            self_type = type(self)
+
+            # If the execution does not return a value
+            if result is None:
```
prashastk: This condition can be moved above self_type = type(self)

blurr/core/field.py
```diff
@@ -100,3 +100,46 @@ def restore(self, snapshot) -> None:
...
+                return self
+
+            # If the method executed is defined in the base type and a base type object is returned
+            # (and not the current type), then wrap the base object into an object of the current type
+            if isinstance(result, self_type.__bases__) and not isinstance(
+                    result, self_type):
+                return self_type(result)
```
prashastk: Hmmm this would create a new copy of the object every time in this case.

Another approach would be to make ComplexTypeBase store a reference to the complex object.
```
class ComplexTypeBase(ABC):
  def __init__(self, complex_object):
     self._complex_object = complex_object
```
Then ```__getattribute__``` called ```self._complex_object.__getattribute__```

What do you think?

indradeep: Adding TODO for it

PR #76, opened 2018-03-28
Pypi deployment and versioning

### LINKED ISSUE #30: PyPI deployment
In order to enable PyPI deployment we need to:
 - Configure circleci to deploy to Pypi on tags
 - Find a way to serialise versioning from the tag, so the CLI shows the correct version
 - Be able to trigger automatic deployments by using Github Release functionality

mserranom:
Fixes #30


setup.py
```diff
@@ -15,13 +33,35 @@ def readme():
...
+        return "0." + os.getenv('CIRCLE_BUILD_NUM')
+
+
+def description():
+    if not is_ci_build():
+        return "LOCAL"
+    elif is_release():
+        return os.getenv('CIRCLE_TAG')
+    else:
+        return "0." + os.getenv('CIRCLE_BUILD_NUM')
```
indradeep: Is this the `blurr-dev` Pypi release?  if so, can we add some more comments along the file for clarity?

mserranom: This is wrong, description() was intended to be different, will edit the whole thing.

PR #78, opened 2018-03-28
Added SparkRunner

prashastk:


blurr/core/record.py
```diff
@@ -20,6 +20,8 @@ def __getattr__(self, name):
         """
         When attributes are not found, None is returned
         """
+        if name.startswith('__') and name.endswith('__'):
+            raise AttributeError
```
prashastk: This is needed so that the record object can be 'pickled'

mserranom: no need to add a text argument?

blurr/runner/spark_runner.py
```diff
@@ -0,0 +1,108 @@
+"""
+Usage:
+    spark_runner.py --raw_data=<files> --stream_dtc=<file> [--window_dtc=<file>] --output_file=<file>
```
mserranom: we were using `-` instead of `-` for argument separators.

mserranom:
what are the next steps? (tests, integrate with the cli, ...)

blurr/core/record.py
prashastk: This gets eaten up internally by Python. Added an argument anyways.

blurr/runner/spark_runner.py
prashastk: Changed to -.

prashastk:
@mserranom -Yep, those two are the next steps. Although, I did not add tests intentionally right now. Need to think a bit about adding spark as a dependency to the tests. @indradeep was suggesting that we put the runners in a separate repo. Lets discuss this in our stand-up.

PR #84, opened 2018-03-29
Enable mypy typings

mserranom:
- Enabled mypy
 - Fixed plenty of typing errors
 - Some other errors (most of them due to bad inheritance) are marked with `# type: ignore`
 - `__init__.py` files are required, otherwise `mypy` won't work


mserranom:
@prashastk 

--> "Overall looks fine but I don't quite understand why some of the ignores are needed". 

Me too, I did some investigation in the ignored ones, but couldn't fix them.

--> "I would recommend to not have any ignore right now and let mypy execution throw warnings. We will address the warnings soon and then start failing on new errors. With all the ignores we will likely just ignore them and won't get down to fixing them."

Not sure if I follow. "Let mypy throw warnings" will cause introducing brand new errors,  wouldn't "won't get down to fixing them" be the final result of doing it?

prashastk:
Lets do the clean up + failing on mypy error before we start the next round of major development work. So, I would suggest to not worry about adding new errors right now for the next week or so.

Good to submit with the ignores removed.

mserranom:
Closing the PR, too many conflicts to merge. will do separately.

PR #87, opened 2018-03-30
Add local transform support to the CLI.

prashastk:
Added local transform support to the CLI.  Also, made validate look for yml files in the current directory if no files are specified.


blurr/cli/util.py
```diff
@@ -0,0 +1,30 @@
...
+import os
+import yaml
+
+from blurr.core.syntax.schema_validator import is_streaming_dtc, is_window_dtc
+
+
+def get_yml_files(path: str = '.') -> List[str]:
+    return [
+        os.path.join(path, file) for file in os.listdir(path)
+        if file.endswith('.yml')
```
indradeep: Can we support `.yaml` extension as well?

blurr/cli/cli.py
```diff
@@ -1,7 +1,14 @@
...
 
 
-def cli(arguments: list, out: Out) -> int:
+def cli(arguments: Dict[str, Any], out: Out) -> int:
     if arguments['validate']:
         return validate_command(arguments['<DTC>'], out)
+    elif arguments['transform']:
+        return transform(arguments['--streaming-dtc'],
+                         arguments['--window-dtc'],
+                         arguments['<raw-json-files>'], out)
```
indradeep: Can we add an else clause that shows `unsupported action` or something like that and then shows help

blurr/cli/validate.py
```diff
@@ -1,22 +1,46 @@
...
 from blurr.core.errors import InvalidSchemaError
 from blurr.core.syntax.schema_validator import validate
-from blurr.util.out import Out
+from blurr.cli.out import Out
+
+
+def validate_command(dtc_files: List[str], out: Optional[Out] = None) -> int:
+    all_files_valid = True
+    if len(dtc_files) == 0:
+        dtc_files = get_yml_files()
```
indradeep: There should be a no DTC YAML files found in directory error?

blurr/core/syntax/schema_validator.py
```diff
@@ -83,10 +83,18 @@ def _validate_streaming(dtc_dict: Dict, name: str) -> None:
...
+
 def validate(dtc_dict: Dict, name="dtc") -> None:
-    if dtc_dict['Type'] == 'ProductML:DTC:Window':
+    if is_window_dtc(dtc_dict):
         _validate_window(dtc_dict, name)
-    elif dtc_dict['Type'] == 'ProductML:DTC:Streaming':
+    elif is_streaming_dtc(dtc_dict):
         _validate_streaming(dtc_dict, name)
     else:
         raise ValueError("Document is not a valid DTC")
```
indradeep: There may be other yaml files in the directory - doesn't mean they are invalid DTCs.  We should only analyze the files that have valid types, and print the validation report for those files scanned.

__main__.py
```diff
@@ -1,18 +1,35 @@
...
+    blurr transform [--streaming-dtc=<file>] [--window-dtc=<file>] <raw-json-files> ...
+
+Commands:
+    validate        Runs syntax validation on the list of DTC files provided. If
+                    no files are provided then all *.dtc files in the current
+                    directory are validated.
+    transform       Runs blurr to process the given raw log file. This command
+                    can be run with the following combinations:
+                    1. No DTC provided - The current directory is searched for
+                    DTCs. First first streaming and the first window DTC are used.
```
vigbk: Typo ?

__main__.py
```diff
@@ -1,18 +1,35 @@
...
+                    2. Only streaming DTC given - Transform outputs the result of
+                    applying the on the raw data file.
```
vigbk: Typo

blurr/cli/util.py
prashastk: Done.

blurr/cli/cli.py
prashastk: Happens automatically with docopt.

blurr/cli/validate.py
prashastk: No output is equivalent. When there is a file this is printed out:
```
$ python __main__.py validate
Running syntax validation on ./stream.yml
document is valid
```

blurr/core/syntax/schema_validator.py
prashastk: What if there is a typo in the type? We shouldn't be ignoring those files....

__main__.py
prashastk: Done.

__main__.py
prashastk: Done.

tests/cli/transform_test.py
```diff
@@ -0,0 +1,68 @@
...
+
+from blurr.cli.out import Out
+from blurr.cli.transform import transform
+
+
+def assert_record_in_ouput(record: Any, out: str) -> None:
+    assert json.dumps(record) in out
+
+
+def test_transform_invalid(capsys) -> None:
```
prashastk: Tried out sysout and syserr capture support from pytest instead of using outstub.

tests/cli/transform_test.py
```diff
@@ -0,0 +1,68 @@
...
+    out, err = capsys.readouterr()
+    assert out == ''
+    assert err == ''
+
+
+def test_transform_only_stream(capsys) -> None:
+    assert transform('tests/data/stream.yml', None, ['tests/data/raw.json'],
+                     Out()) == 0
+    out, err = capsys.readouterr()
+    assert_record_in_ouput(('userA/session/2018-03-07T22:35:31+00:00', {
```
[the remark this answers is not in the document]
prashastk: A session key with no timestamp denotes the 'current' session. All older sessions will have a timestamp. This is needed to manage concurrent new sessions being created so that both update the same item in dynamodb.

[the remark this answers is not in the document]
mserranom: ah, makes sense.

tests/cli/transform_test.py
vigbk: Cool

Issue #74, opened 2018-03-27
Introduce logging


indradeep:
A comprehensive system for logging expression evaluation errors exist, so that during dev and in production runs, one can figure out where errors are occurring.

mserranom:
this should include replacing any `print` statement with a logging framework

PR #88, opened 2018-03-30
End to end tests - Streaming

indradeep:
Complex scenarios of using Blurr


blurr/core/complex_fields.py
```diff
@@ -10,17 +10,19 @@ class Map(ComplexTypeBase, dict):
 
     def set(self, key: Any, value: Any) -> None:
         """ Sets the value of a key to a supplied value """
-        self[key] = value
+        if key is not None:
```
prashastk: I don't think this alone is suffiicient. There are other dict functions that can be used to update the dict and those are not being checking for this. e.g. merge, update, etc.

Same applies to other complex types. I would suggest to not do this check right now and then fix it more comprehensively.

blurr/core/evaluation.py
```diff
@@ -114,6 +114,6 @@ def evaluate(self, evaluation_context: EvaluationContext) -> Any:
         try:
             return eval(self.code_object, evaluation_context.global_context,
                         evaluation_context.local_context)
-
-        except Exception as e:
-            raise ExpressionEvaluationError(e)
+        except:
+            # TODO Log exception
```
prashastk: Create an issue for this as well.

tests/extended/stream.yml
```diff
@@ -0,0 +1,90 @@
...
+      # Check for data changes across events
+      # Uses bracket notation to access data
+      - { Name: build, Type: integer, Value: "source['build']" }
+      # Check for prior state value
+      # Checks for variable usage
+      - { Name: is_paid, Type: boolean, Value: state.is_paid or vars.is_paid }
+      # Tests accessing variables.
+      # Tests exceptional cases when variables don't exist
+      - { Name: os_name, Type: string, Value: "vars.os_version_parts[0]" }
+      - { Name: os_version, Type: string, Value: "vars.os_version_parts[1]" }
```
prashastk: Lets use the same formatting at the fields below?

tests/extended/stream.yml
```diff
@@ -0,0 +1,90 @@
...
+      # Tests access to time fields in a session
+      - Name: start_score
+        Type: integer
+        Value: source.score
+        When: time == session.start_time
+
+      - Name: end_score
+        Type: integer
+        Value: source.score
+        When: time == session.end_time
```
prashastk: Lets also test some compatible casting? float value to integer etc?

I think a more comprehensive test for this casting belongs to field_test.py

blurr/core/complex_fields.py
indradeep: That's fine...they require another dictionary to merge / update.  This is basically to ensure that when values are not present, they don't get added to the complex types.  I disagree with your sentiments on this.

blurr/core/evaluation.py
indradeep: Done.  #74

tests/extended/stream.yml
indradeep: Trying to cast by default.  It raises error if cast fails.

tests/extended/stream.yml
indradeep: Made formatting more consistent.

PR #90, opened 2018-04-02
merge-window-anchor Created WindowAggregate by merging anchor aggregate and window

vigbk:


blurr/core/window_data_group.py
```diff
@@ -0,0 +1,115 @@
...
+    def __init__(self, fully_qualified_name: str,
+                 schema_loader: SchemaLoader) -> None:
+        super().__init__(fully_qualified_name, schema_loader)
+        self.window_value = self._spec[self.ATTRIBUTE_WINDOW_VALUE]
+        self.window_type = self._spec[self.ATTRIBUTE_WINDOW_TYPE]
+        self.source = self.schema_loader.get_schema_object(
+            self._spec[self.ATTRIBUTE_SOURCE])
+
+
+class _Window:
```
indradeep: Why is this needed as a separate object?  Why can this not follow the same pattern as we have in session aggregate?

prashastk: Explained offline to Tito. He is convinced that its needed. Will rename to WindowSource.

PR #119, opened 2018-04-03
fixed find_packages

mserranom:


setup.py
```diff
@@ -46,7 +46,7 @@ def version():
         return "0." + os.getenv('CIRCLE_BUILD_NUM')
 
 
-# emit a VERSION file the CLI can use to check current version
+# emit a VERSION file the CLI can use to cpheck current version
```
prashastk: Typo

setup.py
```diff
@@ -60,7 +60,7 @@ def version():
     author="productml.com",
     author_email="info@productml.com",
     url="https://github.com/productml/blurr",
-    packages=find_packages(include="blurr"),
+    packages=find_packages(),
     data_files=[
```
prashastk: Instead of this ```include_package_data=True``` should work too.

mserranom: where?

prashastk: Replace data_files=[] with include_package_data=True

mserranom: ah

Issue #114, opened 2018-04-03
Include identity in every data group implicitly


indradeep:
The identity should be added to each data group - the identity currently only exists as part of the key.  In we are to dump the payload, the identity is missed.  

The identity will also be required to join records for a user in one row - for instance if we want to create a record that includes state, last week, last month rollups, they need to be joined on identity.

### Specs:

Field Name to be added: `id`
Value: `identity` from global context

The field spec will be injected at DataGroup Base (similar to how schema injection happens for Block Aggregate)

PR #122, opened 2018-04-03
Included `identity` to be returned with every data group

indradeep:
All Data groups now execute and return the identity with the record.
Fixes: #114


tests/core/data_group_schema_test.py
```diff
@@ -31,13 +31,23 @@ class MockDataGroupSchema(DataGroupSchema):
     pass
 
 
+def test_data_group_identity_field(data_group_schema_spec):
```
mserranom: sorry to be repetitive: `test_data_group_schema_contains_identity_field` is a better spec name

blurr/core/block_data_group.py
```diff
@@ -21,18 +21,23 @@ def __init__(self, fully_qualified_name: str,
...
+        # Add new field schema to the schema loader
         for field_schema in predefined_field:
             self.schema_loader.add_schema(field_schema,
                                           self.fully_qualified_name)
 
+        return super().extend_schema(spec)
+
     @staticmethod
-    def _build_predefined_fields_spec(
+    def _build_block_start_end_fields_spec(
```
prashastk: A better name IMO would be _build_time_fields_spec. This is already in the block class, we don't need to mention that in the function as well.

blurr/core/data_group.py
```diff
@@ -41,6 +41,24 @@ def _load_store(self, store_name: str) -> 'Store':
             store_name)
         return self.schema_loader.get_schema_object(store_fq_name)
 
+    def extend_schema(self, spec: Dict[str, Any]):
```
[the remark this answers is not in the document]
indradeep: Unfortunately, i don't think there's anything that indicates that a method is to be overridden.  There are some projects that introduce some assertions, but i don't think it's fruitful.

blurr/core/data_group.py
```diff
@@ -41,6 +41,24 @@ def _load_store(self, store_name: str) -> 'Store':
...
+        """ Injects the identity field """
+
+        # Add new field to the schema spec
+        identity_field = self._build_identity_field_spec()
+        spec[self.ATTRIBUTE_FIELDS].insert(0, identity_field)
+
+        # Add new field schema to the schema loader
```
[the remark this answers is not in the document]
indradeep: Agreed.

blurr/core/data_group.py
```diff
@@ -41,6 +41,24 @@ def _load_store(self, store_name: str) -> 'Store':
...
+        self.schema_loader.add_schema(identity_field,
+                                      self.fully_qualified_name)
+
+        return super().extend_schema(spec)
+
+    @staticmethod
```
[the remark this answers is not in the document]
indradeep: I got rid of the method altogether... it's not necessary.

tests/core/data_group_schema_test.py
indradeep: Agreed. Incorproated.

blurr/core/block_data_group.py
indradeep: Incorporated.

indradeep:
Added types for functions that have been added as well.

PR #131, opened 2018-04-03
Validation to error out on using resrved keywords as fieldnames

### LINKED ISSUE #86: Handle possible name clashes due to  __getattr__
We have to figure out a way to handle name clashes from class properties that are set and fields from the dtc.  Ex:  A field name to set to type in the dtc. When retrieving the value for calculations, you will get the self.type value of that particular class instead.

indradeep:
Created a new constraint for YAML validation to be able to specify exclusion criteria for identifiers.

Added the set of reserved keywords in the streaming and window DTC validation schema to error out on reserved keyword fieldnames.

Fixes #86


tests/core/syntax/schema_validator_test.py
```diff
@@ -86,3 +86,10 @@ def test_invalid_datagroup_has_no_fields():
         dtc_dict = load_example('invalid_datagroup_has_no_fields.yml')
         validate(dtc_dict)
     assert 'DataGroups.0.Fields: Required field missing' in str(err.value)
+
+
+def test_invalid_field_name_reserved_keywords():
```
mserranom: as in https://github.com/productml/blurr/pull/133:

my rule of thumb to have proper test case names: it should have at least a __verb__, otherwise it can't be a spec. For this case, something like:

This could be: `test_reserved_keywords_as_field_names_raises_invalid_schema_error`

indradeep: Incorporated change.

blurr/core/syntax/dtc_streaming_schema.yml
```diff
@@ -33,7 +33,7 @@ data_group_block:
   Fields: list(include('field'), min=1)
 
 field:
-  Name: identifier()
+  Name: identifier(exclude=['name', 'identity', 'needs_evaluation', 'schema', 'snapshot', 'nested_items', 'evaluation_context'])
```
prashastk: I thought we decided to use _name, etc for local variables? Then we just need to ensure that no identified should start with _ as they are supposed to be for internal use.

indradeep:
As per discussion with @prashastk I have renamed all the local fields to start with _ so that they do not conflict with the user-defined fields.  Also, changed the validator to error out on user defined field names that start with _

blurr/core/block_data_group.py
```diff
@@ -37,8 +37,7 @@ def extend_schema(self, spec: Dict[str, Any]) -> Dict[str, Any]:
         return super().extend_schema(spec)
 
     @staticmethod
-    def _build_time_fields_spec(
-            name_in_context: str) -> List[Dict[str, Any]]:
+    def _build_time_fields_spec(name_in_context: str) -> List[Dict[str, Any]]:
         """
         Constructs the spec for predefined fields that are to be included in the master spec prior to schema load
         :param name_in_context: Name of the current object in the context
```
prashastk: Should we change the name of the predefined fields as well and make them start with a _? e.g. _start_time?
What happens if someone creates a start_time field?

blurr/core/syntax/dtc_window_schema.yml
```diff
@@ -32,8 +32,8 @@ data_group_window:
 
 
 field:
-  Name: identifier()
+  Name: identifier(exclude=['name', 'identity', 'needs_evaluation', 'schema', 'snapshot', 'nested_items', 'evaluation_context', 'window_source'])
```
prashastk: Do we still need needs_evaluation etc here?

blurr/core/window_data_group.py
```diff
@@ -56,28 +56,28 @@ def prepare_window(self, start_time: datetime) -> None:
         should be generated.
         """
         # evaluate window first which sets the correct window in the store
-        store = self.schema.source.store
-        if self.schema.window_type == 'day' or self.schema.window_type == 'hour':
+        store = self._schema.source.store
+        if self._schema.window_type == 'day' or self._schema.window_type == 'hour':
             self.window_source.view = self._load_blocks(
```
prashastk: Should this be _window_source as well?

blurr/core/block_data_group.py
indradeep: Agreed.  Changing

blurr/core/syntax/dtc_window_schema.yml
indradeep: Nope, we no longer need these.  Cleaning up.

blurr/core/window_data_group.py
indradeep: Yes

blurr/core/syntax/schema_validator.py
```diff
@@ -41,23 +42,51 @@ def get_name(self) -> str:
...
 
+    def fail(self, value):
+        return '\'%s\' starts with _ or containing whitespace characters.' % value
+
 
 class Expression(Validator):
     TAG = 'expression'
 
+    set_not_allowed = '\'%s\' sets value using `=`.'
+    invalid_python_expression = '\'%s\' is an invalid python expression.'
```
prashastk: Make these two as CAPS as they are pretty much constants here.

tests/core/syntax/schema_validator_test.py
```diff
@@ -78,11 +95,20 @@ def test_invalid_incorrect_expression():
...
 
 def test_invalid_datagroup_has_no_fields():
     with raises(InvalidSchemaError) as err:
         dtc_dict = load_example('invalid_datagroup_has_no_fields.yml')
         validate(dtc_dict)
     assert 'DataGroups.0.Fields: Required field missing' in str(err.value)
+
+
+def test_reserved_keywords_as_field_names_raises_invalid_schema_error():
+    with raises(InvalidSchemaError) as err:
```
prashastk: You can add a match='Name: \'_name\' starts with _ or containing whitespace characters.' here instead of the assert below.

prashastk: Same for other instances in this file.

blurr/core/syntax/schema_validator.py
indradeep: Done.

tests/core/syntax/schema_validator_test.py
indradeep: Done.

PR #133, opened 2018-04-03
String as default field type when field type is missing

### LINKED ISSUE #132: Missing Field Type should default to string
If the `Type` attribute inside a field is not defined, a `string` type should be assumed.

indradeep:
Added a schema extension snippet in DataGroup to inject a 'string' Type for fields with missing type.

Fixes #132

### DIFF
--- a/tests/core/data_group_schema_test.py
+++ b/tests/core/data_group_schema_test.py
@@ -18,6 +18,9 @@ def data_group_schema_spec() -> Dict[str, Any]:
             'Name': 'event_count',
             'Type': 'integer',
             'Value': 5
+        }, {
+            'Name': 'missing_type',
+            'Value': 'test'
         }]
     }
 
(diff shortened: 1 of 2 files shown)


tests/core/data_group_schema_test.py
```diff
@@ -62,3 +65,14 @@ def test_data_group_schema_initialization_without_store(
     name = schema_loader.add_schema(data_group_schema_spec)
     data_group_schema = MockDataGroupSchema(name, schema_loader)
     assert data_group_schema.store is None
+
+
+def test_field_without_type(data_group_schema_spec):
```
mserranom: my rule of thumb to have proper test case names: it should have at least a name, otherwise it can't be a spec.

This could be: `test_field_without_type defaults_to_string_`

indradeep: I've incorporated the change.

tests/core/data_group_schema_test.py
```diff
@@ -62,3 +65,14 @@ def test_data_group_schema_initialization_without_store(
     name = schema_loader.add_schema(data_group_schema_spec)
     data_group_schema = MockDataGroupSchema(name, schema_loader)
     assert data_group_schema.store is None
+
+
+def test_field_without_type_defaults_to_string(data_group_schema_spec):
+    schema_loader = SchemaLoader()
+    del data_group_schema_spec['Store']
```
prashastk: Why is this needed?

indradeep: That's because we don't need to load a store to test our case.   I've change the test data to not require this statement.

PR #141, opened 2018-04-04
Base item collection unit tests

vigbk:


vigbk:
Will add the test for get item after you merge @indradeep

indradeep:
There is a PR that adds support for __getitem__.  However that is not merged to master yet.  I suppose we the order of merge will determine who adds the test cases.  For now this is good.

tests/core/base_item_collection_test.py
```diff
@@ -0,0 +1,200 @@
...
+    This class is to test abstract behavior, and thus, adds no functionality
+    """
+
+    pass
+
+
+def test_evaluate_needs_evaluation_false(
+        collection_schema_spec: Dict[str, Any]) -> None:
+    schema_loader = SchemaLoader()
+    collection_schema_spec['When'] = 'False'
```
prashastk: Lets also write one test where 'When' evaluates to an runtime error.

vigbk: Sure

PR #149, opened 2018-04-04
Ypaf formatting improvements

indradeep:
Created setup.cfg file to add the YAPF configurations there instead of the build file.

Removed style declaration from the build file.

Formatting changes:
- Column Width increased to 100 from default 79
- Enabled splitting complex comprehensions

### DIFF
--- a/blurr/cli/cli.py
+++ b/blurr/cli/cli.py
@@ -14,5 +14,4 @@ def cli(arguments: Dict[str, Any], out: Out) -> int:
             source = arguments['--source'].split(',')
         elif arguments['<raw-json-files>'] is not None:
             source = arguments['<raw-json-files>'].split(',')
-        return transform(arguments['--streaming-dtc'],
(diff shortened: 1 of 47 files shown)


mserranom:
I much prefer the 79 line formatting:
 - It's pretty standard, IDEs setup and viewer setups are by default configured like that
 - More personally, Code Reviews can't be done side by side in a 13" screen in github

indradeep:
I understand the issue here Miguel.  However, here is the rationale for setting character limit to 100.

When you read code on GitHub, if you have a 79 character width, you use a fraction of the screen space.  A wider character width prevents unnecessary wraps - as you can see on about 300 lines.   By increasing the character width, we will improve readability of code on GitHub.  

Outside the side-by-side usecase for reviews, all other use cases and development will benefit from a wider character width.   Will you consider reducing the zoom size on your browser for side-by-side reviews?

PR #158, opened 2018-04-06
Add logging support

prashastk:
- Added a logging.py file to handle logging.
- Removed the Out class.
- Exception handling to crash when unexpected errors happen and silently consume exceptions when they are expected.
- Added a --debug flag to CLI to help debug whats going wrong.


blurr/core/logging.py
```diff
@@ -0,0 +1,83 @@
...
+import os
+import sys
+import threading
+
+# Don't use this directly. Use _get_logger() instead.
+_logger = None
+_logger_lock = threading.Lock()
+
+
+def _get_logger() -> logging.Logger:
```
prashastk: Using this approach from https://github.com/tensorflow/tensorflow/blob/r1.7/tensorflow/python/platform/tf_logging.py

blurr/cli/transform.py
```diff
@@ -1,20 +1,20 @@
...
 
 def transform(stream_dtc_file: Optional[str], window_dtc_file: Optional[str],
-              raw_json_files: List[str], out: Out) -> int:
+              raw_json_files: List[str]) -> int:
     if stream_dtc_file is None and window_dtc_file is None:
         stream_dtc_file, window_dtc_file = get_stream_window_dtc_files(
             get_valid_yml_files(get_yml_files()))
 
     if stream_dtc_file is None:
-        out.eprint('Streaming DTC file not provided and could not be found in '
```
mserranom: I don't think we should use logging here. Explained when requesting review changes.

blurr/__main__.py
```diff
@@ -47,9 +48,18 @@ def read_version(version_file: str) -> str:
         return 'LOCAL'
 
 
+def setup_logs(debug: bool) -> None:
+    logger = logging.getLogger()
+    logger.setLevel(logging.DEBUG if debug else logging.INFO)
+    _handler = logging.StreamHandler(sys.stdout)
+    _handler.setFormatter(logging.Formatter('%(levelname)s - %(message)s'))
+    logger.addHandler(_handler)
```
mserranom: How is logging configured at module level? When I enable `--debug` I'd expect to see blurr logs only, not those of the modules

Is there any internal python magic doing that?

mserranom:
also filed a related issue: https://github.com/productml/blurr/issues/159

blurr/core/evaluation.py
```diff
@@ -102,5 +104,14 @@ def evaluate(self, evaluation_context: EvaluationContext) -> Any:
             return eval(self.code_object, evaluation_context.global_context,
                         evaluation_context.local_context)
         except Exception as err:
-            # TODO Log exception
+            # Evaluation exceptions are expected because of missing fields in the source 'Record'.
+            logging.debug('{} in evaluating expression {}. Error: {}'.format(
```
mserranom: logging.error() ?

blurr/__main__.py
mserranom: nervermind, found logging class!

blurr/core/logging.py
```diff
@@ -0,0 +1,83 @@
...
+            try:
+                # This is only defined in interactive shells.
+                if sys.ps1:
+                    interactive = True
+            except AttributeError:
+                # Even now, we may be in an interactive shell with `python -i`.
+                interactive = sys.flags.interactive
+
+            # If we are in an interactive environment (like Jupyter), set loglevel
+            # to INFO and pipe the output to stdout.
```
mserranom: I think we should be explicit separating logging and output. It's different things, we shouldn't entangle them. Left a top-level comment.

blurr/runner/identity_runner.py
```diff
@@ -77,11 +80,23 @@ def execute_window_dtc(identity: str, schema_loader: SchemaLoader,
...
         if key.group != block_obj._schema.name:
             continue
-        if window_transformer.evaluate_anchor(block_obj.restore(data)):
-            window_data.append(window_transformer.flattened_snapshot)
+        try:
+            blocks += 1
+            if window_transformer.evaluate_anchor(block_obj.restore(data)):
+                anchors += 1
+                window_data.append(window_transformer.flattened_snapshot)
+        except PrepareWindowMissingBlocksError as err:
```
mserranom: ⚠️ logging.error ?

blurr/core/evaluation.py
prashastk: Keeping it debug because of it producing a lot of expected errors. Logging it as errors loses its meaning when a lot of these are expected.

blurr/runner/identity_runner.py
prashastk: This error is thrown by the library and the caller would be expected to deal with it. No need to log it here imo.

blurr/core/logging.py
prashastk: How should this work in an interactive environment then? Where should be log be printed?

Note that this is only applicable if existing handlers are not specified.

tests/core/expression_test.py
```diff
@@ -65,6 +100,43 @@ def test_invalid_expression() -> None:
         Expression(code_string)
 
 
-def test_execution_error() -> None:
+def test_execution_error(caplog) -> None:
```
indradeep: We can perhaps add a test case where if you specify a higher log-level, nothing is printed to the output - just to ensure log levels work as expected.

I'm not sure if it is valuable however -- just a thought.

indradeep:
While at it, can you add logging to `Field.evaluate()`

blurr/runner/identity_runner.py
prashastk: From the identity_runner perspective its expected. From the core library perspective its for the caller to determine what to do with this.

tests/core/expression_test.py
prashastk: Instead added a check in this test itself to check the level of the log being produced.

PR #165, opened 2018-04-07
Add unit tests for the Transformer and StreamingTransformer.

### LINKED ISSUE #154: Add unit tests for Transformer and StreamingTransformer

prashastk:
Fixes #154

### DIFF
--- a/blurr/core/transformer.py
+++ b/blurr/core/transformer.py
@@ -37,13 +37,6 @@ class TransformerSchema(BaseSchemaCollection, ABC):
             for schema_spec in self._spec.get(self.ATTRIBUTE_STORES, [])
         }
 
-        # Load nested schema items
-        self.nested_schema: Dict[str, Type[DataGroup]] = {
-            schema_spec[self.ATTRIBUTE_NAME]: self.schema_loader.get_nested_schema_object(
-                self.fully_qualified_name, schema_spec[self.ATTRIBUTE_NAME])
-            for schema_spec in self._spec[self._nested_item_attribute]
-        }
-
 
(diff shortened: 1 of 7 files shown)


blurr/core/transformer.py
```diff
@@ -37,13 +37,6 @@ def __init__(self, fully_qualified_name: str, schema_loader: SchemaLoader) -> No
             for schema_spec in self._spec.get(self.ATTRIBUTE_STORES, [])
         }
 
-        # Load nested schema items
-        self.nested_schema: Dict[str, Type[DataGroup]] = {
```
prashastk: This was happening in the BaseItemCollection class. Not needed here!

tests/core/streaming_transformer_test.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def test_streaming_transformer_schema_get_identity_from_record(schema_loader: SchemaLoader,
+                                                               schema_spec: Dict[str, Any]) -> None:
+    schema_spec['Identity'] = 'source.user'
+    streaming_dtc = schema_loader.add_schema(schema_spec)
+    transformer_schema = StreamingTransformerSchema(streaming_dtc, schema_loader)
+    assert transformer_schema.get_identity(Context({
+        'source': Record({
+            'user': 'user1'
```
indradeep: This `user1` will not be reflected in as part of the processing right? The literal `user1` will be returned.

Can we call this `user2` instead then?

prashastk: I updated the schema spec in this function to schema_spec['Identity'] = 'source.user'

PR #169, opened 2018-04-09
Use compile for expression validation as opposed to ast.parse.

### LINKED ISSUE #167: YAML validation fails for valid Python expressions that are invalid YAML
This is expected behavior, but falls short of our claim that users can write `any python expression` and is currently breaking the fbt example. 

Examples: 

**Function arguments**

`Value: sorted(last_24_hours.products_bought.extend(next_24_hours.products_bought), key=itemgetter(1), reverse=True)[:3]`

Running `blurr validate` fails with `DataGroups.3.Fields.0.Value: 'sorted(last_24_hours.products_bought.extend(next_24_hours.products_bought), key=itemgetter(1), reverse=True)[:3]' sets value using `=`.` even though the usage of = is correct within the sorted function

**Mapping values**
`Value: sorted(last_24_hours.products_bought.extend(next_24_hours.products_bought), key=lambda x: x[1], reverse==True)[:3]` is another way to write the sorted function but it is not a valid YAML. 

yamllint.com throws a validation error `(<unknown>): mapping values are not allowed in this context at line 82 column 103` - the culprit being `key=lambda x: x[1]`  (Blurr validation fails too but doesn't show the problem). 

There are likely to be many other situations where Python expressions will lead to failed YAML validation - do we need a bigger strategy for how to handle this?

prashastk:
Compile can handle the cases where = is fine and where it would cause an error.

Fixes #167

### DIFF
--- a/tests/core/syntax/schema_validator_test.py
+++ b/tests/core/syntax/schema_validator_test.py
@@ -47,7 +47,7 @@ def load_example(file):
 
 
 def test_validation_errors_contain_dtc_name_and_schema_location():
-    with raises(InvalidSchemaError, match='Error validating data dtc_name with schema') as err:
+    with raises(InvalidSchemaError, match='Error validating data dtc_name with schema'):
         dtc_dict = load_example('invalid_wrong_version.yml')
         validate(dtc_dict, 'dtc_name')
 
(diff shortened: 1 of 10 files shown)


tests/core/syntax/schema_validator_test.py
```diff
@@ -47,7 +47,7 @@ def load_example(file):
 
 
 def test_validation_errors_contain_dtc_name_and_schema_location():
-    with raises(InvalidSchemaError, match='Error validating data dtc_name with schema') as err:
+    with raises(InvalidSchemaError, match='Error validating data dtc_name with schema'):
         dtc_dict = load_example('invalid_wrong_version.yml')
```
indradeep: It seems that throughout the document, you're loading the example and validating.  Why not call it `validate_example` and save the two steps?

prashastk: Done. Convert to a parameterized test.

Issue #70, opened 2018-03-27
Validate example DTCs at build time


mserranom:
Currently we can include anything as a DTC in the `examples` section of the project.

We should invoke `blurr validate` as part of `build.sh`, so we don't end up with broken DTCs in the sources.

PR #171, opened 2018-04-09
Validate example DTCs at build time

mserranom:
fixes https://github.com/productml/blurr/issues/70


build.sh
```diff
@@ -21,12 +21,22 @@ export PYTHONPATH=`pwd`:$PYTHONPATH
 pipenv run pytest -v --cov=blurr
 
 
+echo "validating code from docs..."
+EXAMPLES=( "tutorial/streaming-dtc.yml" "tutorial/window-dtc.yml" \
+           "offer-ai/offer-ai-streaming-dtc.yml" "offer-ai/offer-ai-window-dtc.yml" \
+           "frequently-bought-together/fbt-streaming-dtc.yml" "frequently-bought-together/fbt-window-dtc.yml" )
+for DTC in "${EXAMPLES[@]}"
+do
+	pipenv run python -m blurr validate examples/${DTC}
```
prashastk: Why not do this as part of the unit test infra?

We should also test the running of these examples and not just the validation.

mserranom:
@prashastk moved the test to pytest. Only validation so far because I'm not too sure how to approach a test for the full transformation. Any suggestion? snapshot testing perhaps?

examples/frequently-bought-together/fbt-window-dtc.yml
```diff
@@ -28,6 +28,16 @@ Anchor:
   # We're free to change this the max value to whatever we want, of course
   Max: 1
 
+
+# Window aggregations are stored in S3 (a Store)
+# Defines the Store properties
+Store:
```
prashastk: This is not needed (and it wouldn't do anything right now).

tests/examples/frequently_bought_together_test.py
```diff
@@ -0,0 +1,9 @@
+from blurr.cli.validate import validate_file
+
+
+def test_streaming_dtc_is_valid():
+    assert validate_file('examples/frequently-bought-together/fbt-streaming-dtc.yml') == 0
```
prashastk: How about making this into a parametrized test? (for all the example files).

prashastk:
@mserranom we should have the sample raw data and the expected output of the sample raw data as files in the examples directory. Then the parameterized test takes in a raw data file, dtc file(s) and output file. It would process the raw data using the dtc file(s) and compare it against the expected output.

PR #173, opened 2018-04-09
Add support for reset to reset the state of the items.

prashastk:


mserranom:
can you add some more context on why this is done, the existing problem, ...?

PR #185, opened 2018-04-11
Use __new__ instead of __init__ for immutable objects

vigbk:


vigbk:
First step before changing types. Will update types in a separate PR

tests/core/block_data_group_test.py
```diff
@@ -95,6 +95,7 @@ def test_block_data_group_schema_evaluate_with_split(block_data_group_schema_spe
     time = datetime(2018, 3, 7, 19, 35, 31, 0, timezone.utc)
     block_data_group = create_block_data_group(block_data_group_schema, time, identity)
     block_data_group.evaluate()
+    print(block_data_group)
```
mserranom: to be deleted

blurr/core/base.py
```diff
@@ -71,14 +71,17 @@ class BaseItem(ABC):
     Base class for for all leaf items that do not contain sub-items
     """
 
-    def __init__(self, schema: BaseSchema, evaluation_context: EvaluationContext) -> None:
+    def __new__(cls, schema: BaseSchema, evaluation_context: EvaluationContext, *args,
+                **kwargs) -> None:
```
prashastk: Why do we have *args, **kwargs in some places and not in others?

blurr/core/base.py
```diff
@@ -71,14 +71,17 @@ class BaseItem(ABC):
     Base class for for all leaf items that do not contain sub-items
     """
 
-    def __init__(self, schema: BaseSchema, evaluation_context: EvaluationContext) -> None:
+    def __new__(cls, schema: BaseSchema, evaluation_context: EvaluationContext, *args,
```
prashastk: Should this change be made for the schema classes too?

indradeep:
We are deferring mypy for future.

PR #197, opened 2018-04-15
Added comparison to Spark

roy100:


README.md
```diff
@@ -14,12 +14,16 @@ Blurr transforms `raw data` into `features` for model training and prediction us
...
 >Coming up with features is difficult, time-consuming, requires expert knowledge. 'Applied machine learning' is basically feature engineering
 
 >--- Andrew Ng
 
+# How is Blurr different from Spark/Kafka?
+
+Blurr is a __data transform definition__, it does not store any data.
+
+The idea is to define data transformations in Blurr, and these data transformations can execute on any stream processor like Spark or Kafka. Because real world infrastructure is extremely diverse, Blurr is designed to run on virtually any infrastructure stack.
+
```
mserranom: `Blurr is a __data transform definition__, it does not store any data`

Blurr is not a definition, perhaps the DTC is? (Descriptor is a much better word in my opinion). Also not sure what storage has to do with spark.

A potential rephrase:

```
Blurr is designed around DTCs, which define __how__ data is transformed. 

While Blurr provides its own runners to execute transformations, it is designed to be processor-agnostic, so transformations can be run by different stream processors like Spark or Kafka. 

In fact, because real world infrastructure is extremely diverse, Blurr has been designed from the scratch to run on virtually any infrastructure stack
```

`designed from the scratch to run on virtually any infrastructure stack` <-- is that true? What is/will be our dynamodb/s3 dependency?

PR #201, opened 2018-04-17
Activity block aggregate

indradeep:
The successor of block aggregate that only aggregates time series data partitioned by a period of inactivity.

This version of the activity aggregate only processes sorted events.  If the events are out of order, it will produce inconsistent results.  

The out-of-order time series processing feature will be raised in another PR.

Since this is a milestone PR, the documentation associated with the ActivityAggregate will be created when it is fully feature complete.

This is Part 2 of #186


mserranom:
could you add more info in what this block does? Ideally in the docs as part of the PR

tests/core/aggregate_activity_test.py
```diff
@@ -0,0 +1,84 @@
...
+    return [
+        Record({'id': 'user1', 'event_value': 10, 'event_time': '2018-01-01T01:01:01+00:00'}),
+        Record({'id': 'user1', 'event_value': 100, 'event_time': '2018-01-01T01:01:05+00:00'}),
+        Record({'id': 'user1', 'event_value': 1, 'event_time': '2018-01-01T01:02:01+00:00'}),
+        Record({'id': 'user1', 'event_value': 1000, 'event_time': '2018-01-01T03:01:01+00:00'}),
+        Record({'id': 'user1', 'event_value': 10000, 'event_time': '2018-01-02T01:01:01+00:00'})
+    ]
+
+
+def test_split_by_inactivity(activity_aggregate_schema, activity_events):
```
mserranom: types

blurr/core/aggregate_activity.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def __init__(self, schema: ActivityAggregateSchema, identity: str,
+                 evaluation_context: EvaluationContext) -> None:
+        super().__init__(schema, identity, evaluation_context)
+        self._label = None
+
+    def evaluate(self) -> None:
+        time = self._evaluation_context.global_context['time']
+
+        if self._start_time and (time < self._start_time - self._schema.separation_interval or
```
mserranom: There's a single test case, just seeing an `if` statement like this I would expect much wider coverage.

blurr/core/aggregate_label.py
```diff
@@ -31,19 +29,15 @@ def extend_schema(self, spec: Dict[str, Any]) -> Dict[str, Any]:
 
 
 class LabelAggregate(BlockAggregate):
-    """
-    Manages the aggregates for block based roll-ups of streaming data
-    """
+    """ Aggregates records in blocks by a label calculated from teh record """
```
mserranom: typo `the`

blurr/core/aggregate_label.py
```diff
@@ -31,19 +29,15 @@ def extend_schema(self, spec: Dict[str, Any]) -> Dict[str, Any]:
...
     def __init__(self, schema: LabelAggregateSchema, identity: str,
                  evaluation_context: EvaluationContext) -> None:
         super().__init__(schema, identity, evaluation_context)
         self._label_value = None
 
     def evaluate(self) -> None:
-        """
-        Evaluates the current item
-        """
+        """ Evaluates the current item """
```
mserranom: comments should be only in the parent class.

blurr/core/syntax/dtc_streaming_schema.yml
```diff
@@ -33,6 +33,14 @@ label_aggregate:
   Label: expression(required=True)
   Fields: list(include('field'), min=1)
 
+activity_aggregate:
+  Name: identifier()
+  Type: enum('Blurr:Aggregate:ActivityAggregate')
+  Store: identifier(required=False)
+  When: expression(required=False)
+  SeparateByInactiveSeconds: int(required=True, min=1)
```
mserranom: Shouldn't it be `SplitByInactiveSeconds` to be consistent with the rest of the code? (or use Separate everywhere instead)

tests/core/aggregate_activity_test.py
vigbk: This needs to cover more possibilities.

indradeep: Agreed.

blurr/core/aggregate_label.py
indradeep: Fixed.

blurr/core/aggregate_label.py
indradeep: Done.

blurr/core/syntax/dtc_streaming_schema.yml
indradeep: We are getting rid of the Block Aggregate as it existed earlier. I'll update across the board and introduce the `separate` terminology and remove `split`.  The cleanup will come in in a separate PR that takes out `BlockAggregate` - which will necessitate the replacement across the board.

indradeep:
Added more information on the PR summary.  This edition of the activity aggregate only works with time-sorted timeseries data.  The ability to handle out-of-order events will be coming in another PR.

tests/core/aggregate_activity_test.py
```diff
@@ -82,3 +88,50 @@ def test_split_by_inactivity(activity_aggregate_schema, activity_events):
...
+    assert len(store_state) == 1
+
+    evaluate_event(activity_events[1], activity_aggregate)
+    activity_aggregate.persist()
+
+    store_state = activity_aggregate._schema.store.get_all()
+    assert len(store_state) == 1
+
+
+def test_evaluate_separate_on_inactivity(activity_aggregate_schema: ActivityAggregateSchema,
```
vigbk: What is the difference between this and test_evaluate_split_by_inactivity in terms of what it is trying to test?

indradeep: Great question.   This is the one that actually tests for the separation behavior.  The other one tests for eventual state / expected state.  I've renamed the other method to make that more relevant.

blurr/core/syntax/dtc_streaming_schema.yml
mserranom: sounds good!

tests/core/aggregate_activity_test.py
vigbk: Oh okay. Makes sense.

blurr/core/aggregate_block.py
```diff
@@ -79,3 +80,6 @@ def evaluate(self) -> None:
             self.__init__(self._schema, self._identity, self._evaluation_context)
 
         super().evaluate()
+
+    def persist(self, timestamp=None) -> None:
```
prashastk: typing for timestamp

indradeep: I'll ignore this for now... this interface will change when we implement the store.

Issue #181, opened 2018-04-10
Handle bad data


mserranom:
Feedback Received: "Feature request: having a way to define assertions/validations and what happens when you get bad data"

thanosqr:
Hey Miguel!

So the example in the tutorial,
```
 - Name: games_played
       Type: integer
       When: source.event_id == 'game_start'
       Value: session_stats.games_played + 1

- Name: games_won
       Type: integer
       When: source.event_id == 'game_end' and source.won == '1'
       Value: session_stats.games_won + 1
```

will turn
```

1 { "user_id": "09C1", "session_id": "915D", "country" : "US", "event_id": "game_start" }
2 { "user_id": "09C1", "session_id": "915D", "country" : "US", "event_id": "game_end", "won": 1 }
3 { "user_id": "09C1", "session_id": "915D", "country" : "US", "event_id": "game_start" }
4 { "user_id": "09C1", "session_id": "915D", "country" : "US", "event_id": "game_end", "won": 1 }

```

into `915D    09C1    2    2`

But what happens if row 3 is missing? I'm assuming that games_played will stay 1 and games_won will be 2.  The user might instead want to remove the session from the results or have an error message. Note that in this case we can see that the result is wrong in its end state, so we could catch it by doing a check after processing. But sometimes it can only be detected while processing, e.g. if rows 1 and 4 are missing.

Implementation-wise, my first thought was to be able to define something like:

```
- Name: games_won
       Type: integer
       When: source.event_id == 'game_end' and source.won == '1'
       Value: session_stats.games_won + 1
       AssertBefore: session_stats.games_played > session_stats.games_won
--or
       AssertAfter: session_stats.games_played >= session_stats.games_won
       IfError: ignore session_id
```

But it can get a bit messy. Then I thought, an assert is pretty much the same as `When`, so we could add a field like this:

```
- Name: cannot_win_more_than_played
       Type: integer
       When: session_stats.games_played > session_stats.games_won
       Value: session_stats.cannot_win_more_than_played + 1
```

or perhaps, to avoid have lots of validation fields:
```
- Name: game_validation
       Type: error_map
       When1: session_stats.games_played > session_stats.games_won
       Value1: "fewer games_played than games_won"

       When2: ...
       Value2: ""
```
the error_map type should evaluate to true if there were any errors and have the count of the different errors as entries but it changes the general structure.

[...]

PR #211, opened 2018-04-17
Docs update for bad data handling

roy100:
Captured @thanosqr and @prashastk conversation on 'handling bad data' issue (https://github.com/productml/blurr/issues/181) and turned it into documentation. Hopefully it will be useful


docs/DTC Snippets.md
```diff
@@ -70,3 +70,40 @@ Fields:
...
+```yaml  
+def valid_start_and_end(games_list):
+  # Determine that each game_start has a corresponding game_end.
+  # and return True / False accordingly.
+```
+As more events are processed the value of session_stats.valid_session will maintain whether this session is valid or not. The fields are evaluated in the order they are defined so the valid_session field should be defined after game_start_and_end field.
+
+# Per event validation
+
+Validate individual events for things like missing field values, reuse fields etc. Some ways to do so are:
```
prashastk: This sentence seems broken. How about:
"Some of the ways to validate individual events are:"

docs/DTC Snippets.md
```diff
@@ -70,3 +70,40 @@ Fields:
     Value: [item for sublist in source.products_bought for item in sublist]
     # Flattens the list of list of tuples of products_bought
 ```
+
+# Data validation
+
+# Determine correctness of a session
+
+One possible way of determining the correctness of a session (without relying on the order of events) could be done with:
```
prashastk: Does the rest of the doc mention session? Might be better to call it an Aggregate block?

PR #228, opened 2018-04-20
type-strings-into-constants Extracted blurr types into string constants.

vigbk:
**PR Checklist**

_Please make sure to review and check all of these items:_

**What kind of change does this PR introduce?**

_Describe the big picture of your changes here_

**Other information**

fixes #XYZ


vigbk:
I think we should have some consistent naming in our types. Sometimes we append the type to the end sometimes we don't. eg: Blurr:Transform:Streaming, Blurr:Aggregate:VariableAggregate. What do you guys think?

blurr/core/constants.py
```diff
@@ -0,0 +1,9 @@
+BLURR_TRANSFORM_STREAMING = "Blurr:Transform:Streaming"
+BLURR_TRANSFORM_WINDOW = "Blurr:Transform:Window"
+BLURR_AGGREGATE_BLOCK = "Blurr:Aggregate:BlockAggregate"
```
indradeep: Let us include the change we discussed here - for Aggregates, remove the superfluous `Aggregate` in the end.  Remember to do a mass replace form the documentation too!

blurr/core/constants.py
```diff
@@ -0,0 +1,9 @@
+BLURR_TRANSFORM_STREAMING = "Blurr:Transform:Streaming"
```
prashastk: How about creating a type.py file which defines these as an enum:
```
class Type(Enum):
    BLURR_TRANSFORM_STREAMING = "Blurr:Transform:Streaming",
    BLURR_AGGREGATE_BLOCK = "Blurr:Aggregate:BlockAggregate",
```

Then passing around Types will be clear as an object as opposed to arbitrary strings. This class can also encapsulate other type related functionality that we may want to add in the future.

blurr/core/constants.py
vigbk: Will do

blurr/core/constants.py
vigbk: Sure. Probably Call it DTCType to be more specific.

prashastk: file being in blurr/core makes it about the DTC. don't think we need to specify that (just like we don't call other classes / objects as DTC)

vigbk:
Okay sure

blurr/core/constants.py
vigbk: sure

vigbk:
Updated PR after converting from string constants to enums.
Also removed redundant suffixes from string types eg: Blurr:Aggregate:VariableAggregate -> Blurr:Aggregate:Variable.

I will update the documentation in another commit for ease of review.

Issue #243, opened 2018-04-24
Add mock tests and extended tests for Dynamo DB Store


indradeep:
**I certify I'm including:**

- [x] A descriptive title
- [x] A description of the problem I'm trying to solve, including *why* I think this is a problem
- [x] An overview of the suggested solution

**Problem**
The current Dynamo DB implementation uses Dynamo DB directly on AWS.  This is time consuming and expensive on every run.  

**Solution**
Use moto library to mock Dynamo DB
Use Dynamo Store as part of extended examples to ensure consistency

PR #244, opened 2018-04-24
Dynamo DB Store Implementation

indradeep:
**PR Checklist**

_Please make sure to review and check all of these items:_

- [x] Docs have been added / updated - Not relevant

**What kind of change does this PR introduce?**

Implemented the DynamoDB store with the current Store API specs.  Included test cases to test out the necessary scenarios.

The  tests hit AWS directly.  Need to implement tests using moto so as not to have the AWS dependency

The store has not been incorporated into the rest of the examples.  This should happen with Moto integration.  This has been added to issue #243


prashastk:
Instead of hitting AWS can we set up a local dynamodb server and hit that instead?

blurr/store/dynamo_store.py
```diff
@@ -0,0 +1,134 @@
...
+                        'KeyType': 'RANGE'
+                    },
+                ],
+                ProvisionedThroughput={
+                    'ReadCapacityUnits': self.rcu,
+                    'WriteCapacityUnits': self.wcu
+                }
+            )
+
+        while self.table_name not in self.dynamodb_client.list_tables()['TableNames'] or \
```
vigbk: Can use table.meta.client.get_waiter('table_exists').wait(TableName=self.table_name)  to check if the table exists ? 
http://boto3.readthedocs.io/en/latest/guide/dynamodb.html

blurr/core/store_key.py
```diff
@@ -41,8 +41,8 @@ def __str__(self):
         return Key.PARTITION.join([self.identity, self.group])
 
     def __eq__(self, other: 'Key') -> bool:
-        return (self.identity, self.group, self.timestamp) == (other.identity, other.group,
-                                                               other.timestamp)
+        return False if not other else \
+            (self.identity, self.group, self.timestamp) == (other.identity, other.group, other.timestamp)
```
prashastk: Might be simpler to say `return other and ...`

blurr/store/dynamo_store.py
```diff
@@ -0,0 +1,134 @@
...
+
+        self.dynamodb_client = boto3.client('dynamodb')
+        self.dynamodb_resource = boto3.resource('dynamodb')
+
+        # If table does not exist, create the table
+        if self.table_name not in self.dynamodb_client.list_tables()['TableNames']:
+            self.dynamodb_client.create_table(
+                AttributeDefinitions=[
+                    {
+                        'AttributeName': 'identity',
```
prashastk: Hmmm we already have a _identity in the aggregate. Should we instead call it identity there or use _identity here?

blurr/store/dynamo_store.py
```diff
@@ -0,0 +1,134 @@
...
+                        'AttributeType': 'S'
+                    },
+                    {
+                        'AttributeName': 'dimensions',
```
prashastk: This could lead to some confusion because there wouldn't a 1:1 map between these dimensions and the `Dimensions` defined in IdentityAggregates.

Lets rename this to range_key or sort_key?

blurr/store/dynamo_store.py
```diff
@@ -0,0 +1,134 @@
...
+                ProvisionedThroughput={
+                    'ReadCapacityUnits': self.rcu,
+                    'WriteCapacityUnits': self.wcu
+                }
+            )
+
+        while self.table_name not in self.dynamodb_client.list_tables()['TableNames'] or \
+                self.dynamodb_client.describe_table(TableName=self.table_name).get('Table', {}).get('TableStatus',
+                                                                                                    '') != 'ACTIVE':
+            sleep(1)
```
prashastk: Will this loop for ever if for some reason the table could not be created?

blurr/store/dynamo_store.py
```diff
@@ -0,0 +1,134 @@
...
+
+        self.table = self.dynamodb_resource.Table(self.table_name)
+
+    @staticmethod
+    def dimensions(key: Key):
+        return key.group + key.PARTITION + key.timestamp.isoformat() if key.timestamp else key.group
```
prashastk: Shouldn't it have key.group and key.timestamp? Also perhaps this can move into the `Key` class.

blurr/store/dynamo_store.py
```diff
@@ -0,0 +1,134 @@
...
+
+        def prepare_record(record: Dict[str, Any]) -> Tuple[Key, Any]:
+            dimensions = record['dimensions'].split(Key.PARTITION)
+            key = Key(record['identity'], dimensions[0], parser.parse(dimensions[1]))
+            return key, self.clean(record)
+
+        records = [prepare_record(item) for item in response['Items']] if \
+            'Items' in response else []
+
+        # Ignore the starting record
```
prashastk: Why?

tests/store/dynamo_store_test.py
```diff
@@ -0,0 +1,89 @@
+import time
```
prashastk: Lets also test a more complete example. Take tests/data/stream.yml and update the store in it to use dynamodb and run that as a test?

blurr/core/store_key.py
indradeep: Done

blurr/store/dynamo_store.py
indradeep: We must decouple the storage from the internal representation of fields.  I'd like to keep them isolated.

blurr/store/dynamo_store.py
indradeep: Agreed.  Renamed `identity` to `partition_key` and `dimensions` to `range_key`

blurr/store/dynamo_store.py
indradeep: Explained why.

blurr/store/dynamo_store.py
indradeep: It is not necessary for timestamp to be present always.  The same structure is used for identity storage as well.

It is up to the store to determine how they want the underlying storage to be.  We do not want to dictate that at the `Key` level.  Plus, `Key` will change with the new store.

blurr/store/dynamo_store.py
indradeep: Changed the code to use Vignesh's suggestion - using the Waiter instead.

tests/store/dynamo_store_test.py
indradeep: Yes of course... staged for another PR.

blurr/store/dynamo_store.py
indradeep: Implemented.

blurr/store/dynamo_store.py
```diff
@@ -0,0 +1,131 @@
...
+                ProvisionedThroughput={
+                    'ReadCapacityUnits': self.rcu,
+                    'WriteCapacityUnits': self.wcu
+                }
+            )
+            # Wait until the table creation is complete
+            self.table.meta.client.get_waiter('table_exists').wait(TableName=self.table_name, WaiterConfig={'Delay': 5})
+
+    @staticmethod
+    def dimensions(key: Key):
```
prashastk: Lets rename this method to range_key?

tests/store/dynamo_store_test.py
```diff
@@ -0,0 +1,126 @@
...
+    assert store.get(Key('test_user', 'test_group')) == {'string_field': 'string', 'int_field': 1}
+
+
+def test_save_time(store: DynamoStore) -> None:
+    start_time = datetime(2018, 1, 1, 1, 1, 1, 1, tzinfo=timezone.utc)
+    store.save(Key('test_user', 'test_group', start_time), {'string_field': 'string2', 'int_field': 2})
+    assert store.get(Key('test_user', 'test_group', start_time)) == {'string_field': 'string2', 'int_field': 2}
+
+
+def test_get_range(store: DynamoStore) -> None:
```
prashastk: Lets break this down into smaller tests

PR #256, opened 2018-04-26
Use local DynamoDB for unit tests.

### LINKED ISSUE #243: Add mock tests and extended tests for Dynamo DB Store

prashastk:
#### Pull Request Checklist:

Use local DynamoDB for unit tests.
Broken down the scripts so that the test script can be run independently.

#### Link to issue
fixes #243


build.sh
```diff
@@ -5,21 +5,15 @@ set -e
 echo "checking requirements..."
 pipenv  --version > /dev/null
```
mserranom: to add: `java -version > /dev/null`

scripts/test.sh
```diff
@@ -0,0 +1,23 @@
...
+status="$(ps aux | grep DynamoDBLocal.jar | grep java)"
+if [ ! "$status" ]; then
+    echo "Local DynamoDB not running"
+    exit 1
+fi
+
+echo "running tests..."
+export PYTHONPATH=`pwd`:$SPARK_HOME/python/lib/pyspark.zip:$SPARK_HOME/python/lib/py4j-0.10.4-src.zip:$PYTHONPATH
+pipenv run pytest -v --cov=blurr
+
```
mserranom: Shouldn't be invoking `stop_dynamodb.sh` here?

scripts/start_dynamodb.sh
```diff
@@ -0,0 +1,12 @@
...
+
+status="$(ps aux | grep DynamoDBLocal.jar | grep java)"
+
+if [ "$status" ]; then
+    echo "Local DynamoDB running"
+else
+    echo "Starting local DynamoDB"
+    # 25877 is blurr in T9
+    java -Djava.library.path=${DYNAMODBPATH?"Need to set DYNAMODBPATH"} -jar $DYNAMODBPATH/DynamoDBLocal.jar -port 25877 -sharedDb &
+    sleep 2
```
mserranom: This opens the door to arbitrary failures, would it be possible to poll the endpoint status?

scripts/test.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -e
+
+if [ ! -d "blurr/core" ]; then
```
mserranom: have you considered changing the directory within the script, to keep all scripts consistently usable from `scritps/` dir?

tests/store/dynamo/dynamo_store_test.py
```diff
@@ -19,11 +20,30 @@ def dynamo_store_spec() -> Dict[str, Any]:
     }
 
 
+BOTO3_DYNAMODB_KWARGS = {
```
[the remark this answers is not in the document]
prashastk: Done.

scripts/start_dynamodb.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+status="$(ps aux | grep DynamoDBLocal.jar | grep java)"
+
+if [ "$status" ]; then
+    echo "Local DynamoDB running"
+else
+    echo "Starting local DynamoDB"
+    # 25877 is blurr in T9
```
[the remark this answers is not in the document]
prashastk: Hmmm... i am not a big fan of including such libs in the code directly. 
Two possible approaches are:
- Have a setup_dev_environ.sh script which download the dependencies and sets them up as well.
- Run the tests in our build docker image which has the environment completely set up. This can even be integrated with pycharm.

Also, If we end up using more AWS stuff in tests we should consider using localstack which internally downloads the dynamodb jar when setting it up for the first time.

Good to tackle in a separate PR IMO.

build.sh
prashastk: Added

scripts/test.sh
prashastk: Added

scripts/start_dynamodb.sh
prashastk: Done. Added a poll every 2 seconds for a max of 30 seconds.

scripts/test.sh
prashastk: Then we will have to build logic to detect the directory and change it accordingly? Also if something breaks in the middle the user is left in a different directory than they started in.

I would prefer to keep it this way and just ensure that the script is being run from the root directory.

scripts/start_dynamodb.sh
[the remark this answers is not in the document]
mserranom: raised https://github.com/productml/blurr/issues/273

scripts/test.sh
mserranom: running `sh script.sh` creates a separate session, you won't be moved to a different directory. 

> I would prefer to keep it this way and just ensure that the script is being run from the root directory.

I'm ok with that, but I'd like **all** the scripts to be consistent and be run from the same place.

PR #257, opened 2018-04-27
Perform basic validaton of schema objects in Schema Loader

### LINKED ISSUE #182: Evaluate Yamale as convenient or not for the long term
Right now we have 2 problems with Yamale:

1.  Doesn't produce the output we desire for validation errors.
2.  It doesn't print out lines/columns with the problem

For 1) we need to check how far we can customise the output errors by overriding the Validators.

For 2) we need to check:
a. whether Yamale supports line
b. if it doesn't support lines, whether we can fork and add support with a yaml parser including lines, like [ruamel](http://yaml.readthedocs.io/en/latest/)

indradeep:
#### Pull Request Checklist:
- [X] Docs have been added / updated - Not relevant

Created a separate validation module that contains the validation routines
The validation of `Name` and `Type` happen at the time of adding schema spec to schema loader.  
Injected schema elements are marked so that they are not validated at schema load.

On the side: renamed some variables and function names to more accurately reflect the task.

This PR does not include the collection of errors before throwing.  Coming in a subsequent PR.

#### Link to issue
fixes #182


tests/core/aggregate_schema_test.py
```diff
@@ -36,38 +33,28 @@ class MockAggregateSchema(AggregateSchema):
...
 
 def test_aggregate_schema_initialization_without_store(aggregate_schema_spec):
     schema_loader = SchemaLoader()
-    name = schema_loader.add_schema(aggregate_schema_spec)
+    name = schema_loader.add_schema_spec(aggregate_schema_spec)
     aggregate_schema = MockAggregateSchema(name, schema_loader)
     assert aggregate_schema.store is None
-
-
-def test_field_without_type_defaults_to_string(aggregate_schema_spec):
```
indradeep: Type must be specified ALWAYS.  We will no longer default to string for missing types for clarity and consistency.

tests/core/schema_loader_test.py
```diff
@@ -59,53 +59,30 @@ def nested_schema_spec() -> Dict:
...
 
 
 def test_add_valid_simple_schema_with_parent() -> None:
     schema_loader = SchemaLoader()
 
-    assert schema_loader.add_schema({'Name': 'test'}, 'parent') == 'test'
-    assert schema_loader.get_schema_spec('parent.test') == {'Name': 'test'}
-
-
-def test_add_valid_nested_schema(nested_schema_spec_bad_type: Dict) -> None:
```
indradeep: These tests are no longer relevant since schema loader will fail at adding a schema block that does not contain the basic `Name` and `Type` attributes.

indradeep:
The new PR encapsulates these changes.  The code has changed so much that this PR is no longer valid.

PR #260, opened 2018-04-27
Store schema

prashastk:
#### Pull Request Checklist:

Introduce a StoreSchema abstraction. The schema classes now only contain the schema for the store. The creation of the store happens later on when the item objects are created.

Internal change - No doc update needed.


blurr/core/schema_loader.py
```diff
@@ -56,12 +57,29 @@ def get_schema_object(self, fully_qualified_name: str) -> 'BaseSchema':
...
             if self.ATTRIBUTE_TYPE not in spec:
-                raise InvalidSchemaError('Type not defined in schema')
+                raise InvalidSchemaError(
+                    'Type not defined in schema for {}'.format(fully_qualified_name))
             self._object_cache[fully_qualified_name] = TypeLoader.load_schema(
                 spec[self.ATTRIBUTE_TYPE])(fully_qualified_name, self)
 
         return self._object_cache[fully_qualified_name]
 
+    def get_store(self, fully_qualified_name: str) -> 'Store':
```
mserranom: unit tests missing for `get_store` in `schema_loader_test.py`

prashastk: Added.

blurr/core/schema_loader.py
```diff
@@ -97,17 +115,17 @@ def get_schema_spec(self, fully_qualified_name: str) -> Dict[str, Any]:
         except:
             raise InvalidSchemaError("{} not declared in schema".format(fully_qualified_name))
 
-    def get_schemas_of_type(self, schema_type: Type) -> List[Tuple[str, Dict[str, Any]]]:
+    def get_schemas_of_type(self, schema_types: List[Type]) -> List[Tuple[str, Dict[str, Any]]]:
```
indradeep: I understand that this returns the spec.  Can we call it `get_schema_specs_of_type`.  To me that is clear that this is the spec.   This will make it consistent with a change I have made to schema loader in PR.

blurr/core/type.py
```diff
@@ -33,6 +33,17 @@ def is_type_equal(actual_type: Union[str, 'Type'], expected_type: 'Type') -> boo
...
+    @staticmethod
+    def is_type_in(actual_type: Union[str, 'Type'], expected_type: List['Type']) -> bool:
+        try:
+            return Type(actual_type) in expected_type
+        except ValueError:
+            return False
+
+    @staticmethod
+    def is_store_type(store_type: Union[str, 'Type']) -> bool:
+        return Type.is_type_in(store_type, [Type.BLURR_STORE_MEMORY, Type.BLURR_STORE_DYNAMO])
```
indradeep: If we do our store check like this, we will be constrained to how we add stores to the system.  Even if we do a text prefix match, it limits naming conventions that do not start with blurr (for third party extensions). 

Ideally type checks should happen with the instances of the type names to see if they are derived from the correct base class.

blurr/core/schema_loader.py
```diff
@@ -56,12 +57,29 @@ def get_schema_object(self, fully_qualified_name: str) -> 'BaseSchema':
...
+    def get_store(self, fully_qualified_name: str) -> 'Store':
+        """
+        Used to generate a store object from the given fully_qualified_name.
+        :param fully_qualified_name: The fully qualified name of the store object needed.
+        :return: An initialized store object
+        """
+
+        if fully_qualified_name not in self._store_cache:
+            schema = self.get_schema_object(fully_qualified_name)
+            if not Type.is_store_type(schema.type):
```
indradeep: Please see comment on `is_type` first before reading this.

To be absolutely clear about types, we should instantiate the type (if schema is present), and then check if the type abides by the norm - that it derives from the correct base class.

blurr/runner/runner.py
```diff
@@ -146,16 +146,17 @@ def _execute_window_dtc(self, identity: str, schema_loader: SchemaLoader) -> Lis
...
 
-    def _get_store(self, schema_loader: SchemaLoader) -> Store:
-        store_schemas = schema_loader.get_schemas_of_type(Type.BLURR_STORE_MEMORY)
-        if not store_schemas:
-            store_schemas = schema_loader.get_schemas_of_type(Type.BLURR_STORE_DYNAMO)
-        return schema_loader.get_schema_object(store_schemas[0][0])
+    @staticmethod
+    def _get_store(schema_loader: SchemaLoader) -> Store:
+        store_schemas = schema_loader.get_schemas_of_type(
+            [Type.BLURR_STORE_MEMORY, Type.BLURR_STORE_DYNAMO])
```
indradeep: By doing type checks not by type name, but by base class, we can also make this code pretty leaproof.

blurr/runner/runner.py
```diff
@@ -146,16 +146,17 @@ def _execute_window_dtc(self, identity: str, schema_loader: SchemaLoader) -> Lis
...
+        return schema_loader.get_store(store_schemas[0][0])
 
+    @staticmethod
     def _get_streaming_transformer_schema(
-            self, schema_loader: SchemaLoader) -> StreamingTransformerSchema:
+            schema_loader: SchemaLoader) -> StreamingTransformerSchema:
         streaming_transformer_schema = schema_loader.get_schemas_of_type(
-            Type.BLURR_TRANSFORM_STREAMING)
```
indradeep: The comments above apply here too.

blurr/core/type.py
indradeep: Based on our discussion, this constraint does not seem to be blocking immediate future extensions / extensibility.   We can revisit this if/when we need to.

blurr/core/schema_loader.py
indradeep: Based on our discussion, I'll defere this comment to later.  Not applicable as of now.

blurr/runner/runner.py
indradeep: Suggest: add `get_all_stores` in schema loader to return all stores.

blurr/core/schema_loader.py
```diff
@@ -97,17 +122,18 @@ def get_schema_spec(self, fully_qualified_name: str) -> Dict[str, Any]:
         except:
             raise InvalidSchemaError("{} not declared in schema".format(fully_qualified_name))
 
-    def get_schemas_of_type(self, schema_type: Type) -> List[Tuple[str, Dict[str, Any]]]:
+    def get_schema_specs_of_type(self,
+                                 schema_types: List[Type]) -> List[Tuple[str, Dict[str, Any]]]:
```
indradeep: This is a great candidate for using the `*schema_types: Type` multiple argument setup.  All code remains the same, and calling the method looks cleaner in a single / multiple argument scenario.

indradeep: `List[Tuple[str, Dict[str, Any]]]: ` I don't get why the order is important here... could this just be a dict?

PR #262, opened 2018-04-27
Foundation for schema validation added to scheam loading, and error collection for raising them together

### LINKED ISSUE #254: Spec for revamped validation
#### Issue Checklist:
- [x] Descriptive title

# Problem Description

We've dismissed Yamale as the tool to validate YAML format. We need line information and customised error messages. This document is the spec for the new validation system.

# Solution Proposal

##  Error Message Format

Each error message will print the line number and an excerpt of the content:

Template:
```
<FILE_PATH>:<LINE_NUMBER> -> <BLOCK_PATH>
<ERROR_MESSAGE>
```

Examples: 
```
./stream.yml:8: -> /Time
invalid python expression

./stream.yml:34: -> /Aggregates/Fields[3]
missing required property "Type"
```

## Errors

**Missing property**

```
./stream.yml:34: -> /Aggregates/Fields[3]
missing required property "Type"
```

**Missing block**

```
./stream.yml:0: -> /
missing required block "Aggregates"
```

**Invalid property**

```
./stream.yml:34: -> /Aggregates/Fields[3]/Tyyype
"Tyyype" is not a valid property 
```

**Invalid block**

```
./stream.yml:28: -> /Aggggregates
"Aggggregates" is not a valid block
```

**Invalid Data Type**

```
./stream.yml:36: -> /Aggregates/Fields[3]/Type
"blah" is not a valid data type 
```

**Invalid Identifier**

```
./stream.yml:35: -> /Aggregates/Fields[3]/Name
"n@me" is not a valid identifier
```
**Reserver Keyword**

```
./stream.yml:35: -> /Aggregates/Fields[3]/Name
"n@me" is not a valid identifier
```

**Invalid Python Expression**

Apart from the message, we'll print the output from Python syntax check:

```
./stream.yml:38: -> /Aggregates/Fields[3]/Value
invalid python expression:
   | <PYTHON_OUTPUT
   | <PYTHON_OUTPUT
```

# Impact on current behavior

The refactored validation will be disabled by default until implementation is complete. In the meantime, Yamale validation will be used, and there will be an available flag to use the new implementation.

indradeep:
#### Pull Request Checklist:

This PR lays the foundation for replacing the `yamale` based schema validation with validation built into the schema spec loading routine.   Schema validation occurs in two places:
- `SchemaLoader`: which requires Name and Type to be present for loads to work properly.
- `BaseSchema` and derived classes - where checks against specific schema elements belonging to the derivative is checked.

Created new Error types for different kind of validation failures by deriving from the old `InvalidSchemaError`.  Created an error collector to collect validation errors so that they can be raised together after all validation errors have been throw in the round.  

Replaced the instances of the old `InvalidSchemaError` with GenericSchemaError.  The new `GenericSchemaError` will be systematically replaced by more specific errors in subsequent PRs.

For readability:
- Renamed `SchemaLoader.add_scheam` to `SchemaLoader.add_scheam_spec`
- Renamed `BaseSchema.extend_schema` to `BaseSchema.extend_schema_spec`

#### Link to issue
fixes #254


blurr/core/errors.py
```diff
@@ -1,8 +1,100 @@
...
+        self.attribute = attribute
+
+
+class SchemaErrorCollection:
+    def __init__(self, *args):
+        self.log: Dict[str, List(InvalidSchemaError)] = defaultdict(list)
+        for arg in args:
+            self.add(arg)
+
+    def add(self,
```
mserranom: this class, and this method particular cries for a dedicated unit test!

blurr/core/errors.py
```diff
@@ -1,8 +1,100 @@
...
+    @property
+    def errors(self) -> List[InvalidSchemaError]:
+        return list(chain.from_iterable(self.log.values()))
+
+    @property
+    def has_errors(self) -> bool:
+        return len(self.log) > 0
+
+
+class RequiredAttributeError(InvalidSchemaError):
```
mserranom: Should we moce the Schema Errors to a dedicate module?

blurr/core/base.py
```diff
@@ -21,21 +22,38 @@ def __init__(self, fully_qualified_name: str, schema_loader: SchemaLoader) -> No
...
             self._spec[self.ATTRIBUTE_WHEN]) if self.ATTRIBUTE_WHEN in self._spec else None
-        self.description = self._spec.get(self.ATTRIBUTE_DESCRIPTION, None)
+        self.description: str = self._spec.get(self.ATTRIBUTE_DESCRIPTION, None)
 
-    def extend_schema(self, spec: Dict[str, Any]) -> Dict[str, Any]:
+    def extend_schema_spec(self, spec: Dict[str, Any]) -> Dict[str, Any]:
         """ Extends the defined schema specifications at runtime with defaults """
         return spec
 
+    def validate_required(self, *attributes) -> SchemaErrorCollection:
```
mserranom: My main concern with this approach is taking all the functionality that right now is clearly defined in the yamale-related code and merge it with `*Schema` classes.

I'd much prefer splitting clearly the responsibilities into different components. Perhaps a different class within this file? (same as *Schema and *Item are different)

blurr/core/validator.py
```diff
@@ -0,0 +1,67 @@
...
+                                       InvalidIdentifierError.Reason.STARTS_WITH_UNDERSCORE))
+        elif not value.isidentifier():
+            errors.add(
+                InvalidIdentifierError(fully_qualified_name, spec, attribute,
+                                       InvalidIdentifierError.Reason.INVALID_PYTHON_IDENTIFIER))
+
+    return errors
+
+
+def validate_required(fully_qualified_name: str, spec: Dict[str, Any],
```
mserranom: bad name, required is an adjective, a noun is missing (required_something)

blurr/core/base.py
indradeep: Validation needs to be part of both Schema construction as well as the Item construction.  It does not make sense to decouple validation from these classes.

As you can see the actual validation code resides in the validation module.  Creating signatures in base is only a wrapper that provides some of the basic parameters by default.  

Can you articulate why validation should be separate?  The behavior of the validation is tightly couple with the code itself, and it needs to change if there is any schema change with it.   In the next PR, note that the errors are stored in SchemaLoader, which centrall maintains all the schema objects as well.

blurr/core/validator.py
indradeep: These are validating attributes witin the the schema.  Is there a case where it is validating anything else?

blurr/core/errors.py
indradeep: I think this would happen eventually.  Let errors grow a little more and then separation can happen as part of a subsequent PR.

blurr/core/errors.py
indradeep: Agreed.

blurr/core/base.py
mserranom: I'll wait for the next PR then.

blurr/core/errors.py
mserranom: ok then

blurr/core/validator.py
mserranom: then `validate_required_att` could work. It's just not english when an adjective is hanging. When I see something like this I expect the first argument "filling the void" for the name, but that's not the case.

blurr/core/base.py
```diff
@@ -50,16 +68,21 @@ def __init__(self, fully_qualified_name: str, schema_loader: SchemaLoader,
         :param schema_loader: Schema repository that returns schema spec by fully qualified name
         :param nested_schema_attribute: Name of the attribute that contains the nested elements
         """
+        self._nested_item_attribute = nested_schema_attribute
+
```
prashastk: Why move this up?

blurr/core/base.py
prashastk: `validate_required` as a name doesn't quite tell me what it does. I thought this is checking if validation is required or not but it seems like its checking for required attributes. lets call it validate_required_attributes instead?

blurr/core/base.py
```diff
@@ -21,21 +22,38 @@ def __init__(self, fully_qualified_name: str, schema_loader: SchemaLoader) -> No
...
-    def extend_schema(self, spec: Dict[str, Any]) -> Dict[str, Any]:
+    def extend_schema_spec(self, spec: Dict[str, Any]) -> Dict[str, Any]:
         """ Extends the defined schema specifications at runtime with defaults """
         return spec
 
+    def validate_required(self, *attributes) -> SchemaErrorCollection:
+        """ Validates that the schema contains a series of required attributes """
+        return validate_required(self.fully_qualified_name, self._spec, *attributes)
+
+    def validate_identity(self, *attributes) -> SchemaErrorCollection:
```
prashastk: identity meaning is overloaded here. We already have something that we call identity. I am not quite sure what this is supposed to do.

blurr/core/base.py
```diff
@@ -21,21 +22,38 @@ def __init__(self, fully_qualified_name: str, schema_loader: SchemaLoader) -> No
...
+        """ Validates that a schema attribute can be a python valid identifier """
+        return validate_identifier(self.fully_qualified_name, self._spec, *attributes)
+
+    def validate(self, errors: SchemaErrorCollection) -> SchemaErrorCollection:
```
prashastk: By whom and when is this called?

blurr/core/errors.py
```diff
@@ -1,8 +1,100 @@
...
+from itertools import chain
+from os import linesep
+from typing import List, Dict, Any, Union
+
+
+class GenericSchemaError(Exception):
+    pass
+
+
+class InvalidSchemaError(Exception, ABC):
```
prashastk: What is the difference between `InvalidSchemaError` and `GenericSchemaError`?

blurr/core/validator.py
```diff
@@ -0,0 +1,67 @@
...
+
+from blurr.core.errors import InvalidIdentifierError, RequiredAttributeError, EmptyAttributeError, SchemaErrorCollection
+
+ATTRIBUTE_NAME = 'Name'
+ATTRIBUTE_TYPE = 'Type'
+ATTRIBUTE_INTERNAL = '_Internal'
+
+
+def validate_identifier(fully_qualified_name: str, spec: Dict[str, Any],
+                        *attributes: str) -> SchemaErrorCollection:
```
prashastk: *attributes should be a List[str] right?

blurr/core/validator.py
```diff
@@ -0,0 +1,67 @@
...
+    """ Validates a set of attributes as identifiers in a spec """
+    errors = SchemaErrorCollection()
+    for attribute in attributes:
+        if attribute not in spec:
+            continue
```
prashastk: Should this be an error?

blurr/core/validator.py
```diff
@@ -0,0 +1,67 @@
...
+
+        value = spec[attribute]
```
prashastk: Should this always be str? What if its not?

blurr/core/base.py
```diff
@@ -21,21 +22,38 @@ def __init__(self, fully_qualified_name: str, schema_loader: SchemaLoader) -> No
...
+        """ Contains the validation routines that are to be executed as part of initialization by subclasses"""
```
prashastk: Its not at all clear to me what should be added here v/s what validation is done by SchemaLoader. The comment here needs to be expanded to make that clear.

blurr/core/validator.py
```diff
@@ -0,0 +1,67 @@
...
+
+        if value.startswith('_'):
```
prashastk: Check for ATTRIBUTE_INTERNAL here as well since this function could be called directly as well....

blurr/core/base.py
indradeep: Renamed to `validate_python_identifier_attributes` which hopefully reflects that attributes should be valid python identifiers.  Suggestions?

blurr/core/base.py
indradeep: This is renamed to `validate_schema_spec` in the next PR.  This is called from base init in the next PR.

blurr/core/errors.py
indradeep: `InvalidSchemaError` is the base for all schema errors.  `GenericSchemaError` is the placeholder for Schema Errors that have not been replaced by the new errors.

blurr/core/validator.py
indradeep: It takes in unlimited parameters, and presents itself as a tuple of strings.

blurr/core/validator.py
indradeep: Nope... If an attribute is not present, we ignore.  This is not a requirements checker.  

I specifically wanted to isolate the two.

blurr/core/validator.py
indradeep: This will always be string.  This is managed internally.  I don't see why we need to validate for string again.

blurr/core/base.py
indradeep: This changes in the next PR.  Please ignore for now.

blurr/core/validator.py
indradeep: ATTRIBUTE_INTERNAL starts with an `_`.  Is there any case where we would not do that?

blurr/core/base.py
indradeep: Validate will be executed in the base, which will will in-turn invoke validation in the collection.  The validation for collection checks that the nested items are present.  So, it is essential that nested items are set before the base init is called.

blurr/core/validator.py
```diff
@@ -0,0 +1,86 @@
...
+    errors = SchemaErrorCollection()
+
+    if ATTRIBUTE_INTERNAL not in spec:
+        errors.add([
+            EmptyAttributeError(fully_qualified_name, spec, attribute)
+            for attribute, value in spec.items()
+            if not value
+        ])
+        errors.add(validate_required_attributes(fully_qualified_name, spec, ATTRIBUTE_NAME, ATTRIBUTE_TYPE))
+        errors.add(validate_python_identifier_attributes(fully_qualified_name, spec, ATTRIBUTE_NAME))
```
prashastk: As discussed, lets do:
- Check for ATTRIBUTE_INTERNAL as part of validate_python_identifier_attributes
- Move the empty attribute check to base.py validate_schema_spec implementation.
- Move name and type check to schema_loader directly.

indradeep: Done.

blurr/core/validator.py
```diff
@@ -0,0 +1,86 @@
...
+                              value_type: Union[Type[int], Type[float]] = int,
+                              minimum: Optional[Union[int, float]] = None,
+                              maximum: Optional[Union[int, float]] = None) -> Optional[InvalidNumberError]:
+    if attribute not in spec:
+        return
+
+    try:
+        value = value_type(spec[attribute])
+        if (minimum and value < minimum) or (maximum and value > maximum):
+            raise Exception()
```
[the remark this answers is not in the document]
indradeep: This is just to get execution into the else block.  Changed this to `raise None` to ensure clarity

blurr/core/validator.py
```diff
@@ -0,0 +1,71 @@
...
+    errors: List[InvalidIdentifierError] = []
+
+    checks: List[Tuple[Callable, InvalidIdentifierError.Reason]] = [
+        (lambda x: x.startswith('_'), InvalidIdentifierError.Reason.STARTS_WITH_UNDERSCORE),
+        (lambda x: x.startswith('run_'), InvalidIdentifierError.Reason.STARTS_WITH_RUN),
+        (lambda x: not x.isidentifier(), InvalidIdentifierError.Reason.INVALID_PYTHON_IDENTIFIER),
+    ]
+
+    for attribute in attributes:
+        if attribute not in spec or ATTRIBUTE_INTERNAL in spec:
```
prashastk: Should check spec[ATTRIBUTE_INTERNAL] is True?

tests/core/base_schema_test.py
```diff
@@ -1,54 +1,111 @@
...
-Type: integer
-When: True == True
-'''
-
 
 @fixture
-def test_schema_spec():
-    return yaml.load(TEST_SCHEMA_SPEC)
+def schema_spec():
+    return yaml.load('''
```
prashastk: Lets return dict directly instead of doing a yaml load? Here and below.

tests/core/base_schema_test.py
```diff
@@ -1,54 +1,111 @@
...
+
+def test_schema_collection_missing_nested_attribute_adds_error(
+        schema_collection_spec: Dict[str, Any]):
+    schema_loader = SchemaLoader()
+    name = schema_loader.add_schema_spec(schema_collection_spec)
+    schema = MockSchemaCollection(name, schema_loader, 'MissingNested')
+
+    assert 1 == len(schema.errors)
+    assert isinstance(schema.errors[0], RequiredAttributeError)
+    assert 'MissingNested' == schema.errors[0].attribute
```
prashastk: Everywhere else we are using the pattern of `assert  schema.errors[0].attribute == 'MissingNested'`. Stick to that?

tests/core/schema_loader_test.py
```diff
@@ -60,53 +59,30 @@ def nested_schema_spec() -> Dict:
...
 
 
 def test_add_valid_simple_schema_with_parent() -> None:
     schema_loader = SchemaLoader()
 
-    assert schema_loader.add_schema({'Name': 'test'}, 'parent') == 'test'
-    assert schema_loader.get_schema_spec('parent.test') == {'Name': 'test'}
-
-
-def test_add_valid_nested_schema(nested_schema_spec_bad_type: Dict) -> None:
```
prashastk: Why remove this one?

tests/core/schema_loader_test.py
```diff
@@ -60,53 +59,30 @@ def nested_schema_spec() -> Dict:
...
-    assert schema_loader.add_schema(nested_schema_spec_bad_type) == 'test'
-    assert schema_loader.get_schema_spec('test.test_group') == nested_schema_spec_bad_type[
-        'Aggregates'][0]
-    assert schema_loader.get_schema_spec('test.test_group.country') == nested_schema_spec_bad_type[
-        'Aggregates'][0]['Fields'][0]
-    assert schema_loader.get_schema_spec('test.test_group.events') == nested_schema_spec_bad_type[
-        'Aggregates'][0]['Fields'][1]
-
-
-def test_get_schema_object_error(nested_schema_spec_bad_type: Dict) -> None:
```
prashastk: Why remove this?

tests/core/transformer_streaming_test.py
```diff
@@ -126,9 +128,46 @@ def test_streaming_transformer_evaluate_user_mismatch(schema_loader: SchemaLoade
...
+    streaming_dtc = schema_loader.add_schema_spec(schema_spec)
+    schema = StreamingTransformerSchema(streaming_dtc, schema_loader)
+
+    assert 3 == len(schema.errors)
+    assert isinstance(schema.errors[0], RequiredAttributeError)
+    assert StreamingTransformerSchema.ATTRIBUTE_IDENTITY == schema.errors[0].attribute
+    assert isinstance(schema.errors[1], RequiredAttributeError)
+    assert StreamingTransformerSchema.ATTRIBUTE_TIME == schema.errors[1].attribute
+    assert isinstance(schema.errors[2], RequiredAttributeError)
+    assert StreamingTransformerSchema.ATTRIBUTE_STORES == schema.errors[2].attribute
```
prashastk: Assert pattern.... we have been using assert from_code == expected.

tests/core/transformer_test.py
```diff
@@ -75,18 +68,6 @@ def test_transformer_init(test_transformer) -> None:
     assert test_transformer._evaluation_context.local_context == {}
 
 
-def test_transformer_finalize(test_transformer: MockTransformer,
```
prashastk: Why is this removed?

blurr/core/aggregate_block.py
```diff
@@ -20,18 +19,22 @@ def __init__(self, fully_qualified_name: str, schema_loader: SchemaLoader) -> No
...
+        """ Injects the block start and end times """
+        super().extend_schema_spec()
 
-        # Add new field schema to the schema loader
-        for field_schema in predefined_field:
-            self.schema_loader.add_schema(field_schema, self.fully_qualified_name)
+        if self.ATTRIBUTE_FIELDS in self._spec:
+            # Add new fields to the schema spec
+            predefined_field = self._build_time_fields_spec(self._spec[self.ATTRIBUTE_NAME])
+            self._spec[self.ATTRIBUTE_FIELDS][1:1] = predefined_field
```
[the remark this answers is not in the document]
indradeep: 0:0 is identity which is injected by the previous extension.  Wanted to maintain order.

blurr/core/errors.py
```diff
@@ -1,8 +1,122 @@
...
+
+
+class SchemaErrorCollection:
+    def __init__(self, *args):
+        self.log: Dict[str, List(InvalidSchemaError)] = defaultdict(list)
+        for arg in args:
+            self.add(arg)
+
+    def add(self,
+            item: Union[InvalidSchemaError, List[InvalidSchemaError], 'SchemaErrorCollection']):
```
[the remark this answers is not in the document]
indradeep: Agreed.  Done.

blurr/core/validator.py
indradeep: I don't think it is necessary.  The presence of the attribute is sufficient.  

However, for consistency in behavior, I'm gonna add a True check.

tests/core/base_schema_test.py
indradeep: Sure.

tests/core/base_schema_test.py
indradeep: Agreed.

tests/core/schema_loader_test.py
indradeep: This must have been accidental.  I've reinstated it.

tests/core/schema_loader_test.py
indradeep: This must have been accidental.  I've reinstated it.

tests/core/transformer_streaming_test.py
indradeep: Made consistent with the rest

tests/core/transformer_test.py
indradeep: Moved this to streaming transformer.  Changed the test to use variable aggregate instead for greater coverage.

indradeep:
@mserranom Your rejection was on the grounds that there should be a flag for `Yamale` replacement.  This PR co-exists with `Yamale` and produces no effect on existing behavior.

PR #264, opened 2018-04-28
Schema validation across project

### LINKED ISSUE #254: Spec for revamped validation

indradeep:
#### Pull Request Checklist:
- [x] Docs have been added / updated - N/A

Instead of storing errors in the schema objects, piped all errors to the schema loader.  Added add/get APIs to schema loader to interact with errors.   Added add/get error methods to schema for easy testing and easy error addition.  This has led to the simplification of the validation signatures

Added explicit validations for:

- Field
- Aggregate
- Block Aggregate
- Activity Aggregate
- Anchor
- Window Aggregate
- Transformer
- Streaming Transformer
- Window Transformer
- Store

#### Link to issue
fixes #254


blurr/core/aggregate_activity.py
```diff
@@ -1,17 +1,29 @@
...
+
     ATTRIBUTE_SEPARATE_BY_INACTIVE_SECONDS = 'SeparateByInactiveSeconds'
 
     def __init__(self, fully_qualified_name: str, schema_loader: SchemaLoader) -> None:
         super().__init__(fully_qualified_name, schema_loader)
 
-        self.separation_interval: timedelta = timedelta(
-            seconds=int(self._spec[self.ATTRIBUTE_SEPARATE_BY_INACTIVE_SECONDS]))
+        try:
+            value = int(self._spec[self.ATTRIBUTE_SEPARATE_BY_INACTIVE_SECONDS])
```
mserranom: How can an exception be thrown here?

blurr/core/aggregate_activity.py
```diff
@@ -5,13 +5,21 @@
...
+    """ Aggregates activity into blocks by partitioning around periods of inactivity """
+
     ATTRIBUTE_SEPARATE_BY_INACTIVE_SECONDS = 'SeparateByInactiveSeconds'
 
     def __init__(self, fully_qualified_name: str, schema_loader: SchemaLoader) -> None:
         super().__init__(fully_qualified_name, schema_loader)
 
-        self.separation_interval: timedelta = timedelta(
-            seconds=int(self._spec[self.ATTRIBUTE_SEPARATE_BY_INACTIVE_SECONDS]))
+        if not self.errors:
```
prashastk: This feels weird. Why is this being checked? Is it to prevent the code crashing here if the attribute is not present?

if that is the case then `self._spec.get(self.ATTRIBUTE_SEPARATE_BY_INACTIVE_SECONDS, 0)` would be preferred.

prashastk:
Can we merge this to validation-with-schema and then review it there? It seems like our reviews in that other PR end up being useless when things get overwritten in this subsequent PR.

Try to have functionality implemented in a PR that is not changing right away otherwise there is no point in us reviewing it. For e.g. the first PR should have only introduced the error classes and maybe error checking SchemaLoader without touching the Schema classes.

blurr/core/aggregate_activity.py
indradeep: Changed it to `.get`

PR #267, opened 2018-04-29
Extend schema spec refactoring

### LINKED ISSUE #266: Refactor code to match extend_schema_spec with the patteren set by validate_schema_spec
#### Issue Checklist:
- [x] Descriptive title

### Problem Description
The current function design for `extend_schema_spec` expects the spec to be passed to it for it to modify and return, even though the spec is already available in the current object.

### Solution Proposal
`extend_schema_spec` should follow the same pattern as `validate_schema_spec` which works on the data in the current object.  This reduces the amount of redundant variable passing and returning.

While at it, rename `_spec` in schema to `spec` without the `_` because there is no need to hide the spec in the schema.

None

indradeep:
#### Pull Request Checklist:

`extend_schema_spec` signature was changed to work on the spec in the object

`_spec` in schema was renamed to `spec` to make it consistent with the rest of the variables declared in schema.

#### Link to issue
fixes #266


mserranom:
Approved. Not sure why a PR is need since it's not merging to master, I'd have just merged to the branch straight away. 
I'm a bit worried about that branch BTW, it's looking huge already :/

blurr/core/base.py
```diff
@@ -24,22 +24,22 @@ def __init__(self, fully_qualified_name: str, schema_loader: SchemaLoader) -> No
         """
         self.schema_loader: SchemaLoader = schema_loader
         self.fully_qualified_name: str = fully_qualified_name
-        self._spec: Dict[str, Any] = self.schema_loader.get_schema_spec(self.fully_qualified_name)
+        self.spec: Dict[str, Any] = self.schema_loader.get_schema_spec(self.fully_qualified_name)
```
prashastk: Why make this change? The `_spec` seems like an internal thing for the schemas. You wouldn't want to have that accessed from the BaseItem objects...

blurr/core/base.py
```diff
@@ -24,22 +24,22 @@ def __init__(self, fully_qualified_name: str, schema_loader: SchemaLoader) -> No
...
+        self.type: str = self.spec[self.ATTRIBUTE_TYPE]
 
         self.when: Expression = Expression(
-            self._spec[self.ATTRIBUTE_WHEN]) if self.ATTRIBUTE_WHEN in self._spec else None
-        self.description: str = self._spec.get(self.ATTRIBUTE_DESCRIPTION, None)
+            self.spec[self.ATTRIBUTE_WHEN]) if self.ATTRIBUTE_WHEN in self.spec else None
+        self.description: str = self.spec.get(self.ATTRIBUTE_DESCRIPTION, None)
 
-    def extend_schema_spec(self, spec: Dict[str, Any]) -> Dict[str, Any]:
+    def extend_schema_spec(self) -> None:
```
prashastk: Why is this change needed (given that it was this before and you changed it to take in the spec for modification in https://github.com/productml/blurr/commit/c28d4bd7a1d534ea4de19d79d98acd25bae967e9)?

indradeep: This is code redundant - By doing validation, I realized that taking in the same parameters as in self currently and passing it back is only noise in the signature.   This is cleaner.

blurr/core/base.py
indradeep: So is everything else.  Spec is referred to from the items as well.

indradeep: Reverted to `_spec`

PR #272, opened 2018-04-30
Prefix BaseItem class members

### LINKED ISSUE #233: Prefix a _ to all methods in item classes
**I certify I'm including:**

- [x] A descriptive title
- [ ] If the feature changes current behaviour, reasons why my solution is better

Currently if an aggregate defines a field called `execute` (or any of the other methods we have in the item classes) the evaluation code will return the` execute method` as opposed to the `execute field`. To prevent this kind of a problem all methods and variables should start with _ (even if they are not private) so as the completely separate out the namespace of internal names.

We already don't support `Field` names starting with _ for this very problem.

prashastk:
#### Pull Request Checklist:

Prefix BaseItem class members with `_` for internal use and `run_` for external use. 
Also, stopped inhering from BaseItem for the Anchor class.

#### Link to issue
fixes #233


blurr/runner/local_runner.py
```diff
@@ -44,7 +44,8 @@ def get_identity_records_from_json_files(
         identity_records = defaultdict(list)
         for file in local_json_files:
```
indradeep: Passing thought, in cases like these, I use the initial `f` as part of the for loop as an index, and use the term `file` as the object i'm dealing with in the rest of the code.  Makes the function call parameters more readable :)

tests/core/base_item_test.py
```diff
@@ -67,3 +69,37 @@ def test_base_item_filter_missing(schema_spec: Dict[str, Any]) -> None:
     test_item = get_test_item(schema_spec)
 
     assert test_item._needs_evaluation
+
+
+def test_base_item_method_naming():
```
indradeep: Ultra cool!

blurr/runner/local_runner.py
prashastk: renamed f to file_stream. That should make it much more clear.

tests/core/base_item_test.py
prashastk: I know! I was quite pleased with this myself :)

PR #289, opened 2018-05-04
Adding Expression instanciation errors to ScheamLoader error collection

### LINKED ISSUE #159: Split Exception types: processing schema and runtime
Right now we have one python file where error granularity is defined.

We should make explicit a separation:

* Exceptions raised while processing the schema (such as when a window dtc refers to a non-existing datagroup in a streaming schema)
* Exceptions raised while executing the transformation

The reason is that both errors are handled differently:

The 1st type of error is the user's fault. Something's wrong with the input, and there should be an explicit, properly formatted, tested and reviewed error message to be displayed.

The 2nd is an internal error. If there are no errors of the 1st type, 2nd errors shouldn't be produced. A full stack trace should be reported in this case.

indradeep:
#### Pull Request Checklist:
- [ ] Docs have been added / updated - N/A

Errors raised during expression construction in the schema objects was raised immediately.  

- Changed `InvalidExpressionError` to inherit from `InvalidSchemaError`
- Built a helper in the base to perform null / empty checks and add any raised error to the error collection
- Removed exception wrapping in the `Expression` construction

#### Link to issue
fixes #159

### DIFF
--- a/blurr/core/base.py
+++ b/blurr/core/base.py
@@ -41,7 +41,7 @@ class BaseSchema(ABC):
         """ Builds an expression object.  Adds an error if expression creation has errors. """
 
         expression_string = self._spec.get(attribute, None)
-        if expression_string and not str(expression_string).isspace():
+        if expression_string:
             try:
                 return Expression(str(expression_string))
             except Exception as err:
--- a/blurr/core/base.py
+++ b/blurr/core/base.py
@@ -1,7 +1,7 @@
 from abc import ABC, abstractmethod
 from typing import Dict, Any, Type, TypeVar, Union, List, Optional
 
-from blurr.core.errors import SnapshotError, InvalidSchemaError
+from blurr.core.errors import SnapshotError, InvalidSchemaError, InvalidExpressionError
 from blurr.core.evaluation import Expression, EvaluationContext
 from blurr.core.schema_loader import SchemaLoader
 from blurr.core.store_key import Key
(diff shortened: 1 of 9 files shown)


blurr/core/base.py
```diff
@@ -76,6 +82,11 @@ def validate_schema_spec(self) -> None:
             validate_python_identifier_attributes(self.fully_qualified_name, self._spec,
                                                   self.ATTRIBUTE_NAME))
 
+    def extend_schema_spec(self) -> None:
```
indradeep: Moved this method down - makes more sense as the code reads.

blurr/core/evaluation.py
```diff
@@ -119,10 +119,7 @@ def __init__(self, code_string: str,
         self.code_string = 'None' if not code_string or code_string.isspace() else code_string
         self.type = expression_type
 
-        try:
```
indradeep: There is no need to wrap exceptions here them here because they will be handled and wrapped by the caller.

tests/core/base_schema_test.py
```diff
@@ -59,9 +60,22 @@ def test_base_schema_validate_schema_spec_missing_type_and_empty_when(
...
     assert len(schema.errors) == 2
     assert isinstance(schema.errors[0], EmptyAttributeError)
-    assert schema.errors[0].attribute == 'When'
+    assert schema.errors[0].attribute == BaseSchema.ATTRIBUTE_WHEN
     assert isinstance(schema.errors[1], InvalidIdentifierError)
-    assert schema.errors[1].attribute == 'Name'
+    assert schema.errors[1].attribute == BaseSchema.ATTRIBUTE_NAME
+
+
+def test_base_schema_build_expression_adds_error_on_invalid_expression(schema_spec: Dict[str, Any]):
```
mserranom: maybe we can remove `base_schema` from the test names, since it's already part of the file name?

blurr/core/base.py
```diff
@@ -34,14 +34,20 @@ def __init__(self, fully_qualified_name: str, schema_loader: SchemaLoader) -> No
...
 
-    def extend_schema_spec(self) -> None:
-        """ Extends the defined schema specifications at runtime with defaults. When this method is being extended,
-        the first line should always be:  ```super().extend_schema_spec()``` """
-        pass
+    def build_expression(self, attribute: str) -> Optional[Expression]:
+        """ Builds an expression object.  Adds an error if expression creation has errors. """
+
+        expression_string = self._spec.get(attribute, None)
+        if expression_string and not str(expression_string).isspace():
```
prashastk: Shouldnt `' '` be treated as an `InvalidExpressionError`?

tests/core/base_schema_test.py
indradeep: Done.

blurr/core/base.py
indradeep: Good point.  I took out the filters from both the expression initialization sequence and from the Expression building sequence to ensure that all empty / space, and other invalid expressions raise an `InvalidExpressionError`.

blurr/core/evaluation.py
```diff
@@ -112,17 +112,10 @@ def __init__(self, code_string: str,
         :param code_string: Python code statement
         """
 
-        # For non string single value expressions. Ex: 5, False.
-        code_string = str(code_string)
-
-        # For None / empty code strings
-        self.code_string = 'None' if not code_string or code_string.isspace() else code_string
```
indradeep: This is no longer necessary, since invalid expressions should raise error.

PR #290, opened 2018-05-08
Runner changes - load old state, output both stream dtc & window dtc results, documentation!

### LINKED ISSUE #237: Update documentation for SparkRunner
**I certify I'm including:**

- [x] A descriptive title
- [x] If the feature changes current behaviour, reasons why my solution is better

prashastk:
#### Pull Request Checklist:

- Added support to load old state so that the same raw data need not be processed again and again.
- Output both stream dtc & window dtc results so that the old state can be persisted.
- Added documentation!

#### Link to issue
fixes #237


blurr/runner/local_runner.py
```diff
@@ -29,12 +29,9 @@ def _validate_dtc_syntax(self) -> None:
 
     def _execute_for_all_identities(
             self, identity_records: Dict[str, List[Tuple[datetime, Record]]]) -> None:
```
vigbk: maybe we should create an alias for this List[Tuple[datetime, Record]] ?

prashastk: Renamed Tuple[datetime, Record] to TimeAndRecord.

blurr/runner/runner.py
```diff
@@ -20,22 +18,27 @@
...
+    """
+    An abstract class that provides functionality to:
+    - Convert raw events in Records
+    - Process a list of Records for a user.
+
+    A class that inherits from Runner should do the following:
+    1. Call `get_per_identity_records()` using an iterator of the events available. This returns
+        a generator which creates a Tuple[Identity, TimeAndRecord]] output.
+    2. The Tuple[Identity, TimeAndRecord]] output should be grouped together by the
+         Identity to create a List of TimeAndRecord] per identity.
```
mserranom: here and in the next line, the closing square bracket of `TimeAndRecord]` is hanging.

blurr/runner/spark_runner.py
```diff
@@ -35,44 +36,142 @@ def get_spark_session(spark_session: Optional['SparkSession']) -> 'SparkSession'
...
+        :param old_state_rdd: A previous streaming DTC state RDD as Tuple[Identity, Streaming DTC
+            State]
+        :return: RDD[Identity, Tuple[Streaming DTC State, List of Window DTC output]]
+        """
+        identity_records_with_state = identity_records
+        if old_state_rdd:
+            identity_records_with_state = identity_records.fullOuterJoin(old_state_rdd)
+        return identity_records_with_state.map(lambda x: self._execute_per_identity_records(x))
 
     def get_record_rdd_from_json_files(self,
```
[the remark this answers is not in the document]
prashastk: They do have a dependency to the DTC saved as instance members. The identity and datetime is determined in the record using the stream DTC.

blurr/runner/runner.py
prashastk: Fixed.

blurr/runner/spark_runner.py
```diff
@@ -35,44 +36,142 @@ def get_spark_session(spark_session: Optional['SparkSession']) -> 'SparkSession'
...
+    def print_output(self, per_identity_data: 'RDD') -> None:
+        """
+        Basic helper function to write data to stdout. If window DTC was provided then the window
+        DTC output is written, otherwise, the streaming DTC output is written to stdout.
+
+        WARNING - For large datasets this will be extremely slow.
+
+        :param per_identity_data: Output of the `execute()` call.
+        """
+        if not self._window_dtc:
```
indradeep: It's constructing documents line-by-line - why store it in memory and print - just print straight for optimal memory loading.

prashastk: Discussed offline. print in lambda will be executed in the nodes which we don't want.

Issue #273, opened 2018-05-01
Remove external dependencies from build system


mserranom:
#### Issue Checklist:
- [x] Descriptive title
- [x] Completed sections below

### Problem Description
Right now we have external dependencies that compromise build portability. So far it is:

- Spark
- DynamoDB.jar

#### Spark

For Spark: I still don't understand why we don't have a dependency on `pyspark` in our `Pipenv` when we have explicit spark imports like in [here](https://github.com/productml/blurr/blob/master/blurr/runner/spark_runner.py). 

I understand we don't want the core library to depend on spark, but right now it is what it is, we shouldn't just break the module by not providing all explicit dependencies. The way to address it should be to create an specific issue to split modules and separate some sort of `blurr-spark` which would depend on `blurr`.

@prashastk any insight? we went briefly through this in other PR.

#### DynamoDB.jar

I requested the Jar to be included in the repo. Adding libs is not a good practice, but expecting a manual setup of the library is worse. Options are:

1. Just include the lib in the repo .
2. Provide a docker environment to run builds, and always run local builds with docker.

My suggestion is to do **1)** right now, and [once workflows are available in circleci](https://discuss.circleci.com/t/command-line-support-for-workflows/14510), move to **2)**. This will fix current situation, and in the future we'll be able to run our builds using circleci CLI and leverage features like parallelisation.

### Impact on current behavior

Portability issues to be solved.

mserranom:
@vigbk just to clarify, when we talk about separate folders, we also mean **separate pypi** modules, right?

prashastk:
@mserranom The last time i tried pip install pyspark it didn't work for me without setting some other environment variables. If it works fine then adding it to the dev-packages sounds good to me.

For DynamoDB.jar how much just downloading it if not present as part of the start_dynamodb.sh script?

PR #291, opened 2018-05-09
dynamodb is downloaded instead of provided

mserranom:
#### Pull Request Checklist:

Dynamo is now downloaded instead of expected to be provided by the developer

#### Link to issue
#273


scripts/start_dynamodb.sh
```diff
@@ -1,17 +1,38 @@
...
+    else
+       echo "$DYNAMO_JAR not found, downloading from $DYNAMO_URL"
+       wget $DYNAMO_URL -O $TMP_FILE
+       unzip -o -d $TMP_DIR $TMP_FILE
+    fi
+}
+
+
 check_status
 if [ "$status" ]; then
```
indradeep: What's the scope of the running instance of Dynamo like?  Is for the duration of the container lifecycle? Or the execution cycle?

mserranom: execution cycle I think

PR #292, opened 2018-05-10
Integrated Streaming DTC validation with CLI

indradeep:
#### Pull Request Checklist:
- [X] Tests for the changes have been added

- Formatter for printing the errors raised during schema creation
- Integrated new validation to CLI (for streaming only)
- Added new validator for value within enumeration


blurr/cli/validate.py
```diff
@@ -24,22 +25,38 @@ def validate_file(dtc_file: str) -> int:
...
+            # streaming DTC to be loaded in the schema loader.  Refactoring required.
+            validate_window(dtc_dict)
         print('Document is valid')
         return 0
     except yaml.YAMLError as err:
         eprint('Invalid yaml')
         eprint(str(err))
         return 1
-    except InvalidSchemaError as err:
+    except BaseSchemaError as err:
```
indradeep: This is no longer required.  Keeping for legacy - to be removed in the next pass.

mserranom: a comment indicating that would be useful

blurr/cli/validate.py
```diff
@@ -24,22 +25,38 @@ def validate_file(dtc_file: str) -> int:
     print('Running syntax validation on {}'.format(dtc_file))
     try:
         dtc_dict = yaml.safe_load(open(dtc_file, 'r', encoding='utf-8'))
-        validate(dtc_dict)
+        if is_streaming_dtc(dtc_dict):
+            validate(dtc_dict)
+        else:
+            # TODO: Window DTC validation using the new validation technique requires
+            # streaming DTC to be loaded in the schema loader.  Refactoring required.
+            validate_window(dtc_dict)
```
mserranom: both `validate` methods for streaming and window should be analogous, but the streaming one is in this file, the window one is separate. Is there any way we can achieve the same level of abstraction for both?

tests/core/error_test.py
```diff
@@ -48,10 +49,33 @@ def test_schema_error_collection_merge(spec, errors):
...
+    assert err.value.errors == collection
+
+
+def test_schema_error_collection_formatter(errors):
+    collection = SchemaErrorCollection(errors)
+    formatter = SchemaErrorCollectionFormatter(line_separator='\n')
+
+    err = formatter.format(collection)
+    assert err == '''
+fqn
```
mserranom: merge error?

blurr/cli/validate.py
```diff
@@ -24,22 +25,38 @@ def validate_file(dtc_file: str) -> int:
     print('Running syntax validation on {}'.format(dtc_file))
```
prashastk: We are doing more than just syntax validation now, right?

blurr/cli/validate.py
```diff
@@ -24,22 +25,38 @@ def validate_file(dtc_file: str) -> int:
...
         logging.error(exception_value)
-        eprint('There was an error parsing the document')
+        eprint('There was an error parsing the document. Error:\n' + str(err))
         return 1
 
 
+def validate(spec: Dict[str, Any]) -> None:
+    schema_loader = SchemaLoader()
+    stream_dtc_name = schema_loader.add_schema_spec(spec)
+    schema_loader.raise_errors()
```
prashastk: Calling raise_errors feels weird. Shouldn't `add_schema_spec` and `get_schema_object` just raise errors directly in case of errors?

blurr/cli/validate.py
indradeep: Done.

blurr/cli/validate.py
indradeep: I'm working on this.  Will be available in the subsequent PRs.

blurr/cli/validate.py
indradeep: Nope... it's deeper than that. Change the text to take out `syntax`

blurr/cli/validate.py
indradeep: I'm going to have to change that to unify the validation behavior for Stream/Window DTC.  I realize that there will be runtime errors raised when the Window DTC's streaming dependenices are not met.  However that should not happen at the schema object construction phase - since it changes the whole validation pattern.  

The next PR addresses that in greater detail.

blurr/cli/validate.py
mserranom: awesome!

blurr/core/base.py
```diff
@@ -1,12 +1,12 @@
...
+from typing import Dict, Any, Type, TypeVar, Union, List, Optional, Set
 
-from blurr.core.errors import SnapshotError, InvalidSchemaError, InvalidExpressionError
+from blurr.core.errors import SnapshotError, BaseSchemaError, InvalidExpressionError
 from blurr.core.evaluation import Expression, EvaluationContext
 from blurr.core.schema_loader import SchemaLoader
 from blurr.core.store_key import Key
 from blurr.core.validator import validate_required_attributes, validate_python_identifier_attributes, \
-    validate_number_attribute, validate_empty_attributes
+    validate_number_attribute, validate_empty_attributes, validate_enum_attribute
```
vigbk: What about making these static functions of a Validator class ? Then you can just import Validator ?

blurr/core/validator.py
```diff
@@ -69,3 +71,14 @@ def validate_number_attribute(
     except:
```
vigbk: This doesn't take into account if the minimum or maximum is zero.

tests/core/error_test.py
indradeep: The merge is fine and so are the tests... may have been an issue then... fixed now.

blurr/core/base.py
indradeep: There is no need for a class here.  I could simply import the module which will have the same effect.  The explicit imports are for clarity.

blurr/core/validator.py
indradeep: Minimum = 0, then exception raised on -max to -1
Maximum = 0, then exception raised when 1 to max

Minimum & Maximum = 0, then 0 is the only value allowed.

blurr/core/base.py
vigbk: Sure.

blurr/core/validator.py
vigbk: I tried this code when the minimum is zero, the maximum is 3 and value was -2. This doesn't throw an error.

tests/core/error_test.py
mserranom: so this fqn is a github error?

blurr/core/validator.py
indradeep: This happened because of the way I was checking if maximum / minimum is defined.  These need to be explicit None checks otherwise 0  is the same as unset.

I've added a whole series of tests to try out different scenarios.  I guess the test was missed out... wonderful catch :)

vigbk: Cool.

PR #293, opened 2018-05-11
Updated README with new mybinder urls

mserranom:
#### Pull Request Checklist:
- [x] Tests for the changes have been added . (N/A)

Current MyBinder links in README.md are broken. MyBinder tries to initialise `setup.py` after processing `requirements.txt`. Since it doesn't exist, `setup.py` fails due to dependencies.

Raised an issue in MyBinder: https://github.com/jupyterhub/binderhub/issues/559 

I pushed a [binder-examples](https://github.com/productml/blurr/tree/binder-examples) branch, so examples will live there for the time being.

Seems to be much faster to load than the `master` branch, so we may consider to give this branch a permanent status, or perhaps its own repo.


vigbk:
i think a different repo would be a good idea.

mserranom:
This is not intended to be merged, once reviewed I'll delete the PR

mserranom:
this has changed to the new repo, will delete and raise new one

PR #295, opened 2018-05-12
nyse-example Revert to using native datetime library for faster perfo…

vigbk:
…rmance

#### Pull Request Checklist:

Revert to using native datetime parsing which is faster than the generic parse function from the python-dateutils library

### DIFF
--- a/docs/examples/nyse/nyse-streaming-dtc.yml
+++ b/docs/examples/nyse/nyse-streaming-dtc.yml
@@ -4,11 +4,11 @@ Description: New York Store Exchange Transformations
 Name: nyse
 
 Import:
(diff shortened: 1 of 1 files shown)


vigbk:
I did some profiling for this and the native datetime is in order of magnitude faster.
```python
for i in range(50000):
      new_date = parse("2010-01-04 16:13:57")
    # new_date = datetime.datetime.strptime("2010-01-04 16:13:57", "%Y-%m-%d %H:%M:%S")
```
Running the above code took 7.880 seconds for parse and 0.847 seconds for datetime

prashastk:
Remember to update blurr-examples too.

PR #297, opened 2018-05-15
Schema Loader Generic Errors to Specific Schema Errors

indradeep:
#### Pull Request Checklist:

The Schema Loader used `GenericSchemaError` at several places that did not qualify the kinds of error that were happening.

- Created new error types for schema loads / spec loads / store loads
- Errors are not raised immediately, and instead are added to the error collection
- Found a way to break dependency of WindowDTC on StreamingDTC - validation can now work on isolated files like they used to.


tests/core/schema_loader_test.py
```diff
@@ -140,28 +151,18 @@ def test_get_transformer_name() -> None:
 
 
 def test_get_store_error_not_declared(schema_loader: SchemaLoader):
-    with raises(GenericSchemaError, match='test.memstore not declared in schema'):
```
indradeep: The three test cases now hit the same path - and therefore are redundant.

blurr/core/errors.py
```diff
@@ -89,7 +102,43 @@ def __str__(self):
...
+                 type_class_name: str = None,
+                 expected_base_type: BaseTypes = None,
+                 *args,
+                 **kwargs):
+        super().__init__(fully_qualified_name, spec, attribute, *args, **kwargs)
+        self.reason = reason
+        self.type_class_name = type_class_name
+        self.expected_base_type = expected_base_type
+
+    def __str__(self):
```
prashastk: Should add a unit test for this. Seems complicated enough to make sure that the str format looks expected for the various reasons.

PR #299, opened 2018-05-15
Caching Types for faster loading

### LINKED ISSUE #298: Cache Type classes while loading the various types
#### Issue Checklist:
- [ ] Descriptive title
- [ ] Completed sections below

### Problem Description
The import library is invoked every time a type is loaded which degrades performance.

### Solution Proposal
The Type classes should be cached so as to reduce the number of calls to the import library.

None

indradeep:
Added type caching against fully qualified type names

#### Pull Request Checklist:
- [ ] Tests for the changes have been added - N/A
- [ ] Docs have been added / updated - N/A

Every type instantiation was using the importlib to import type by full name.  This is an expensive operation.

Added a type cache against fully qualified type name - so that the type can be returned immediately.

#### Link to issue
fixes #298

### DIFF
--- a/blurr/core/loader.py
+++ b/blurr/core/loader.py
@@ -1,5 +1,5 @@
 import importlib
-from typing import Any, Union
+from typing import Any, Union, Dict, Optional
 
 from blurr.core.errors import GenericSchemaError
 from blurr.core.type import Type
@@ -53,6 +53,7 @@ SCHEMA_MAP = {
 
 # TODO Build dynamic type loader from a central configuration rather than reading a static dictionary
 
+_class_cache: Dict[str, Type] = {}
 
 class TypeLoader:
     @staticmethod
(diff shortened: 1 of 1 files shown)


vigbk:
Maybe can try profiling both methods with and without caching to see the performance difference.

indradeep:
@mserranom I agree with your general opinion on singletons.  This case however is that of loading static data.  We are not managing types, but instances.  Unless we have a scenario whereby we are dynamically changing python modules at runtime, the type is tied to the package deployed, and does not change unless the package is updated.

In our specific case, the Type cache should be at the application level, and should not change with the jobs.  I do not see an issue here.

prashastk:
flying-by comment - @indradeep I agree with the general sentiment that we should profile this before changing things for the sake of optimizing.

PR #300, opened 2018-05-15
Duplicate errors should not appear twice

indradeep:
#### Pull Request Checklist:
- [] Docs have been added / updated - N/A

We were internally storing the errors raised in a list which means that depending on the paths traversed, an error could get added more than once.

To reconcile this:

- Used a `set` to store errors
- Added `__eq__` and `__hash__` to ensure error object comparisons work as expected
- Since order is no longer maintained in sets, changed the unit tests to look for errors in all errors available instead of referring to them by index.

#### Link to issue
fixes #issue-number


indradeep:
@mserranom Added a test case to check for duplicate resolution checking.

tests/core/error_test.py
```diff
@@ -72,10 +72,23 @@ def test_schema_error_collection_formatter(errors):
...
---> Attribute `empty` under `fqn` cannot be left empty.
---> `identity: _some thing` in section `fqn` is invalid. Identifiers starting with underscore `_` are reserved.
-'''
+===''')
+    assert '--> Attribute `missing` must be present under `fqn`.\n' in err
+    assert '--> Attribute `empty` under `fqn` cannot be left empty.\n' in err
+    assert '--> `identity: _some thing` in section `fqn` is invalid. Identifiers starting with underscore `_` are reserved.' in err
+
+
+def test_schema_error_collection_add_duplicate(errors):
```
indradeep: @mserranom Added as per feedback.

PR #306, opened 2018-05-24
Transitioning to new Validation - Part 1

indradeep:
#### Pull Request Checklist:

This PR aims to cut over validation from Yamale to the new implementation.

It operates at the validation step used by the CLI.

This PR does not remove the old schema validation.

#### Link to issue
fixes #issue-number


blurr/core/anchor.py
```diff
@@ -24,6 +24,7 @@ def __init__(self, fully_qualified_name: str, schema_loader: SchemaLoader) -> No
 
     def validate_schema_spec(self) -> None:
         self.validate_required_attributes(self.ATTRIBUTE_CONDITION, self.ATTRIBUTE_MAX)
```
prashastk: MAX is not required. It not being present means there is no limit.

docs/examples/tutorial/tutorial2-window-dtc.yml
```diff
@@ -6,6 +6,7 @@ SourceDTC: sessions
 
 Anchor:
   Condition: sessions.session_stats.boost == True
+  Max: 1
```
prashastk: Should work without this.

tests/cli/cli_validate_test.py
```diff
@@ -49,3 +61,37 @@ def test_invalid_dtc(capsys):
...
+     {InvalidIdentifierError('example_name.user._name', {}, BaseSchema.ATTRIBUTE_NAME,
+                             InvalidIdentifierError.Reason.STARTS_WITH_UNDERSCORE)}),
+    ('invalid_dtc_type.yml',
+     {InvalidTypeError('ProductMLExample', {}, BaseSchema.ATTRIBUTE_TYPE,
+                       InvalidTypeError.Reason.TYPE_NOT_DEFINED)}),
+    ('invalid_dimension_type.yml',
+     {InvalidValueError('example_name.user.value', {}, BaseSchema.ATTRIBUTE_TYPE,
+                        {Type.INTEGER.value, Type.STRING.value})}),
+])
+def test_invalid_schema(file: str, errors: Set[BaseSchemaError]) -> None:
```
prashastk: I think the CLI test should check the string output as opposed to the error being thrown.

tests/core/syntax/schema_validator_test.py
```diff
@@ -62,21 +62,3 @@ def test_valid_basic_streaming_dtc():
 def test_valid_basic_window_dtc():
     dtc_dict = load_example('valid_basic_window.yml')
     validate(dtc_dict)
-
-
-@mark.parametrize("test_file,err_string", [
-    ('invalid_wrong_version.yml', 'Version: \'2088-03-01\' not in \(\'2018-03-01\',\)'),
```
prashastk: Delete these files from tests/core/syntax/dtcs?

indradeep: Removals are part of the subsequent PR.

docs/examples/tutorial/tutorial2-window-dtc.yml
indradeep: Yup.  Understood.

blurr/core/anchor.py
indradeep: Makes sense.

tests/cli/cli_validate_test.py
indradeep: CLI does not contain the logic of representing an error as a string.  The test ensures that the correct error is thrown.  The error serialization `__str__()` method actually stringifies it.  

Errors tests contain the string validation.

blurr/cli/validate.py
```diff
@@ -53,9 +43,12 @@ def validate_file(dtc_file: str) -> int:
 
 def validate(spec: Dict[str, Any]) -> None:
     schema_loader = SchemaLoader()
-    stream_dtc_name = schema_loader.add_schema_spec(spec)
+    dtc_name = schema_loader.add_schema_spec(spec)
+    if not dtc_name:
+        raise InvalidSpecError(spec)
```
mserranom: Isn't `InvalidSpecError ` handled by `validate_command` ?

indradeep: This error is only raised if a YAML that does not contain `Type` and `Name` attributes is added to the schema loader.  Since the block cannot be added, it returns an invalid `dtc_name`.  

This check cannot be done anywhere else but at the point where the first schema addition happens into the schema loader.  

As far as the translation to known errors happen, All these errors have a `__str__()` defined in the class.  The string construction out of the error happens automatically when raised.

PR #309, opened 2018-05-25
Added context manager to handle adding and clearing of context in transforms

### LINKED ISSUE #199: Use context manager to create a clear scope when evaluating records
Right now we following the pattern of:
```
add record to context
evaluation code
del record from context
```

Instead we can do:
```
with record in context do:
  evaluation code
```

vigbk:
#### Pull Request Checklist:
- [ ] Tests for the changes have been added

Added a scoped context maanger for evaluation context to handle adding and clearing of context when executing a transform.

#### Link to issue
fixes #199

### DIFF
--- a/blurr/core/aggregate_activity.py
+++ b/blurr/core/aggregate_activity.py
@@ -3,6 +3,7 @@ from datetime import timedelta, datetime
 from blurr.core.aggregate_block import BlockAggregate, BlockAggregateSchema
 from blurr.core.schema_loader import SchemaLoader
 from blurr.core.store_key import Key
+from blurr.core.evaluation import EvaluationContext
 
 
 class ActivityAggregateSchema(BlockAggregateSchema):
@@ -26,8 +27,9 @@ class ActivityAggregateSchema(BlockAggregateSchema):
 class ActivityAggregate(BlockAggregate):
     """ Aggregates activity in blocks separated by periods of inactivity"""
 
-    def run_evaluate(self) -> None:
-        time = self._evaluation_context.global_context['time']
+    def run_evaluate(self, evaluation_context: EvaluationContext = None) -> None:
+        evaluation_context = evaluation_context if evaluation_context else self._evaluation_context
+        time = evaluation_context.global_context['time']
 
         if not self._start_time:
             self._load_old_state()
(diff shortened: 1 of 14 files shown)


blurr/core/transformer_streaming.py
```diff
@@ -25,31 +26,43 @@ def validate_schema_spec(self) -> None:
...
         :return: The evaluated identity
         :raises: IdentityError if identity cannot be determined.
         """
-        context = self.schema_context.context
-        context.add_record(record)
-        identity = self.identity.evaluate(context)
+        identity = None
+        context = evaluation_context if evaluation_context else self.schema_context.context
+
+        if record is not None:
```
prashastk: Why not use ScopedEvaluationContext here as (and in get_time) as well?

blurr/core/transformer_streaming.py
```diff
@@ -65,18 +78,19 @@ def run_evaluate(self, record: Record):
...
-        # Add source record and time to the global context
-        self._evaluation_context.add_record(record)
-        self._evaluation_context.global_add('time',
-                                            self._schema.time.evaluate(self._evaluation_context))
-        super().run_evaluate()
-
-        # Cleanup source and time form the context
-        self._evaluation_context.remove_record()
-        self._evaluation_context.global_remove('time')
+        with ScopedEvaluationContext(self._evaluation_context, Context({
```
prashastk: Lets make this as call of EvaluationContext? 

Something like `with self._evaluation_context.scoped_context(Context({})) as ...`

blurr/core/transformer_streaming.py
```diff
@@ -65,18 +78,19 @@ def run_evaluate(self, record: Record):
...
+        })) as evaluation_context:
+            record_identity = self._schema.get_identity(None, evaluation_context)
+            if self._identity != record_identity:
+                raise IdentityError(
+                    'Identity in transformer ({}) and new record ({}) do not match'.format(
+                        self._identity, record_identity))
+
+            # Add source record and time to the global context
+
+            evaluation_context.global_add('time', self._schema.time.evaluate(evaluation_context))
```
prashastk: time can also be added to the scoped context?

blurr/core/aggregate_activity.py
```diff
@@ -38,7 +40,7 @@ def run_evaluate(self) -> None:
             self._persist(self._start_time)
             self.run_reset()
 
-        super().run_evaluate()
+        super().run_evaluate(evaluation_context=evaluation_context)
```
indradeep: Should we consider making evaluation context an explicit optional field in base `run_evaluate`?

blurr/core/transformer_streaming.py
indradeep: Should we consider adding special methods to add source, time and identity to the contexts.  It seems like we are doing it all the time.

PR #310, opened 2018-05-28
Add timestamp and dimension key type.

prashastk:
#### Pull Request Checklist:

Added a TIMESTAMP and DIMENSION KeyType to lay the ground work of making the Key class aware of the type of key that is being used.

For timestamp keys, time is a part of the sort key generated but for dimension keys time is not a part of the sort key. This is so that the item stored against the key can be retrieved with just the dimensions. Time based queries on dimension keys would have to be done on a secondary index or would require a partition + group scan.

Changing Block aggregate to dimensions instead of split condition will be in the next PR.

### DIFF
--- a/blurr/core/aggregate_activity.py
+++ b/blurr/core/aggregate_activity.py
@@ -2,7 +2,7 @@ from datetime import timedelta, datetime
 
 from blurr.core.aggregate_block import BlockAggregate, BlockAggregateSchema
 from blurr.core.schema_loader import SchemaLoader
-from blurr.core.store_key import Key
+from blurr.core.store_key import Key, KeyType
 
 
 class ActivityAggregateSchema(BlockAggregateSchema):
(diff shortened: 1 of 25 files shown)


blurr/core/aggregate_activity.py
```diff
@@ -35,16 +35,17 @@ def run_evaluate(self) -> None:
...
-            self._persist(self._start_time)
+            self._persist()
             self.run_reset()
 
         super().run_evaluate()
 
     def _load_old_state(self):
         most_recent_block = self._store.get_range(
-            Key(self._identity, self._name, datetime.utcnow()), None, -1)
+            Key(KeyType.TIMESTAMP, self._identity, self._name, [], datetime.utcnow()), None, -1)
```
indradeep: The parameters are hard to read - find meanings to for this signature.  It will benefit from parameterized attributes. 

Comment left on Key.

blurr/core/store_key.py
```diff
@@ -1,15 +1,29 @@
...
 class Key:
     """
     A record in the store is identified by a key
     """
     PARTITION = '/'
-
-    def __init__(self, identity: str, group: str, timestamp: datetime = None) -> None:
+    DIMENSION_PARTITION = ':'
+
+    def __init__(self,
```
indradeep: I'd suggest the following:
```python
def __init__(self,  key_type: KeyType, identity: str, group: str, *, dimensions: List[Any] = list(), timestamp: datetime = None) -> None: 
```

The `*,` placed after group will force parameterized on the two key attributes tied to the key type.

blurr/core/aggregate_identity.py
```diff
@@ -90,15 +90,12 @@ def _compare_dimensions_to_fields(self) -> bool:
...
-                       (':').join([str(item.value) for item in self._dimension_fields.values()]),
-                       timestamp)
+        return Key(KeyType.DIMENSION, self._identity, self._name,
+                   [str(item.value) for item in self._dimension_fields.values()])
 
-        return Key(self._identity, self._name, timestamp)
-
-    def _persist(self, timestamp=None) -> None:
+    def _persist(self) -> None:
         # TODO Refactor keys when refactoring store
```
indradeep: This is the refactoring we have been waiting for.  You can remove this TODO

blurr/core/aggregate_activity.py
prashastk: Left a TODO in the code for it.

blurr/core/store_key.py
prashastk: Left a TODO in the code for it.

blurr/core/aggregate_identity.py
prashastk: Done.

PR #319, opened 2018-06-04
Exactly once semantics

### LINKED ISSUE #303: Implement exactly once semantics
TODO: Fill details below

#### Issue Checklist:
- [ ] Descriptive title
- [ ] Completed sections below

### Problem Description
_A description of the problem, including **why** this is a problem.  If it is a bug report, include elevant debug logs / reproduction information_

### Solution Proposal
_An overview of the suggested solution_

_Specify **None** if there isn't any_

_If there is, a description of the impact and **why** the proposed solution is better_

indradeep:
#### Pull Request Checklist:

Blurr should process each record only.  If the record is replayed, Blurr should ignore processing the record.

This has been implemented by doing the following:

1. A new `BloomFilterField` type has been added to retain a bloom filter with all processed Ids
2. A new implicit `Attribute` ahs been added to `Attribute` base that evaluates the current record identity to see if it has been processed.  If it is found to have been processed, the record execution is ignored
3. `StreamingTransformer` now features a `RecordIdentifier` Attribute up top for the developer to specify which filed(s) to use to determine uniqueness.  If `RecordIdentifier` is not defined, the whole record is used.

#### Link to issue
fixes #303


blurr/core/field_complex.py
```diff
@@ -82,3 +86,37 @@ def encoder(value: Any) -> Set:
     @staticmethod
     def decoder(value: Any) -> Set:
         return Set(value)
+
+
+class BloomFilter(ComplexTypeBase, ScalableBloomFilter):
```
prashastk: I don't like using a probabilistic cache for this. You can no longer give exactly once semantics. By definition this works with an expected error rate with which this is 'mostly once semantics'.

tests/core/aggregate_test.py
```diff
@@ -56,51 +58,94 @@ def aggregate_schema_without_store():
...
+    })
+
+    aggregate._evaluation_context.global_add('source', record)
+    aggregate._evaluation_context.global_add('time', parser.parse(record.event_time))
+    aggregate._evaluation_context.global_add('_record_id', record)
+
+    aggregate.run_evaluate()
+    aggregate.run_evaluate()
+
+    assert aggregate.event_count == 2
```
prashastk: Also verify the state of _processed_tracker here and above.

blurr/core/field_complex.py
prashastk: This will likely cause problems with identity aggregate which accumulate stuff over the user's lifetime. We likely need a time component here and data that is older than some threshold.

blurr/core/aggregate.py
```diff
@@ -22,33 +22,58 @@ class AggregateSchema(BaseSchemaCollection, ABC):
...
         """ Injects the identity field """
         super().extend_schema_spec()
 
-        identity_field = {
-            'Name': '_identity',
-            'Type': BtsType.STRING,
-            'Value': 'identity',
-            ATTRIBUTE_INTERNAL: True
-        }
+        self._spec[self.ATTRIBUTE_IS_PROCESSED] = '_record_id in {aggregate}.{tracker} if "_record_id" in globals() or "_record_id" in locals() else False'.format(
```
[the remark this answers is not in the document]
indradeep: _record_id is added implicitly by the streaming transformer if it is not specified explicitly.  See `StreamingTransformer` for details.

